PR #23, opened 2021-01-09
Calculations, Custom Directory, and Graceful Shutdown

robweber:
These are some small updates that don't change the functionality of the program in any meaningful way, but set the stage for some further enhancements (USB drive support, running as a service). There are 3 commits I'll try and summarize below. 

### Extract Info From Video File for calculations

The program already extracted the frame count using ffmpeg.probe, this adds runtime and frames per second as well. Rather than hardcoding the fps it will use whatever is in the video file. This avoids errors when the end of video files are reached and the math doesn't line up to the frame counts. These are stored in a dict that can be used elsewhere when needed. The runtime is unused but again lays some ground work for things like a future timecode display. 

### Added Support for Custom Directory and File loading order. 

The first is just the addition of a ```-D``` or ```--dir``` flag to set the location of the videos folder. I also simplified and re-ordered the file loading code. This was based on my intrepretation of how it should work as I noticed in some situations it didn't work properly. One example being if you change the directory but didn't give a file it tried to load a file that didn't exist in the directory. I also modified the ```movieList``` variable as a one-liner to just load .mp4 files in the directory at start so we don't have to weed them out later. The file loading order now is something like this: 

1. Load ```movieList``` and get all valid .mp4 files into an array
2. If ```-f``` flag is given, load this video no questions asked. If this video doesn't exist in the directory, go to step 3
3. If ```-f``` not given, try loading a video from the ```nowPlaying``` file. Check that it exists in the directory
4. If all else fails load a random file (0 index) from the ```movieList``` variable
5. Log and write this file as ```nowPlaying```

Lines [124](https://github.com/robweber/SlowMovie/blob/v0.2-revisions/slowmovie.py#L124) - [170](https://github.com/robweber/SlowMovie/blob/v0.2-revisions/slowmovie.py#L170) are the ones with these changes

### Added Signal Handler support

I believe this is part of one of the other PRs as well so if you want me to strip this out I can in favor of that one. This adds a handler for when the program is triggered to exit so that the display can be closed out properly. It seems a little counterintuitive but it first initializes the display, then closes it. I noticed a RunTimeError being thrown without the init if the sign was in sleep mode and couldn't be properly closed out. This allows for a graceful shutdown.

### DIFF
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -15,6 +15,7 @@ from PIL import Image
 import ffmpeg
 import fnmatch
 import argparse
+import signal
 
 # Ensure this is the correct import for your particular screen
 from waveshare_epd import epd7in5_V2 as epd_driver
@@ -22,6 +23,15 @@ from waveshare_epd import epd7in5_V2 as epd_driver
 fileTypes = [".mp4", ".mkv"]
 HOME_DIR = os.path.dirname(os.path.realpath(__file__))
 NOW_PLAYING = os.path.join(HOME_DIR, 'nowPlaying')
+
+# function to handle when the is killed and exit gracefully
+def signal_handler(signum, frame):
+    print('Exiting Program')
+
+    # must init to get gpio pins setup correctly and then exit properly
+    epd_driver.epdconfig.module_init()
+    epd_driver.epdconfig.module_exit()
+    exit(0)
 
 def generate_frame(in_filename, out_filename, time):
     try:
@@ -80,6 +90,10 @@ parser.add_argument('-i', '--inc',  default=4,
 parser.add_argument('-s', '--start',
     help="Start at a specific frame")
 args = parser.parse_args()
+
+# add hooks for interrupt signal
+signal.signal(signal.SIGTERM, signal_handler)
+signal.signal(signal.SIGINT, signal_handler)
 
 # set paths to log and video directories
 viddir = os.path.join(HOME_DIR, 'Videos/')
@@ -230,8 +244,3 @@ while 1:
     epd.sleep()
     time.sleep(frameDelay)
     epd.init()
-
-epd.sleep()
-
-epd7in5.epdconfig.module_exit()
-exit()
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -80,7 +80,7 @@ parser = argparse.ArgumentParser(description='SlowMovie Settings')
 parser.add_argument('-r', '--random', action='store_true',
     help="Random mode: chooses a random frame every refresh")
 parser.add_argument('-f', '--file', type=check_mp4,
-    help="Add a filename to start playing a specific film. Otherwise will pick a random file, and will move to another film randomly afterwards.")
+    help="Add a filename to start playing a specific film. Otherwise will pick a random file, and will move to another film randomly afterwards. This option is ignored if nowPlaying file exists.")
 parser.add_argument('-D', '--dir', type=check_dir,
     help='Set the directory that contains the video files. Default is ./Videos/')
 parser.add_argument('-d', '--delay',  default=120,
@@ -115,11 +115,6 @@ if args.random:
 else:
     print ("In play-through mode")
 
-if args.file:
-    print('Try to start playing %s' %args.file)
-else:
-    print ("Continue playing existing file")
-
 # get all mp4 files in video directory
 movieList = sorted(fnmatch.filter(os.listdir(viddir), '*.mp4'))
 print (movieList)
@@ -145,9 +140,9 @@ if(os.path.exists(NOW_PLAYING)):
             nowPlayingFile = line.strip()
         f.close()
 
-        if(nowPlayingFile in viddir):
+        if(nowPlayingFile in movieList):
             currentVideo = nowPlayingFile
-            print('Found now playing file %s' % currentVideo)
+            print('Continue playing existing file %s' % currentVideo)
     except Exception as e:
         print(e)
         pass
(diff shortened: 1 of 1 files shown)


robweber:
Upon further review I did find one edge case for the file loading that the current implementation and my new one don't handle very well. 

Take the example of an execution that takes the ```-f``` parameter. Imagine it runs through that file and starts the next video file in the directory. This updates the ```nowPlaying``` file. At this point there is a reboot to the system or the program is stopped and restarted. 

__Scenerio 1__ under the current implementation the contents of the ```nowPlaying``` file trump the ```-f``` flag. This could be problematic if the user has changed the value of the ```-f``` flag and now their chosen video won't play as the one in the file takes precedence. 

__Scenerio 2__ with my new ordering implementation the ```-f``` option trumps the ```nowPlaying``` value. This has a different issue in that if the video has move on to another in the directory it will reset back to the file given upon a restart. 

What are your thoughts on the correct handling of this type of thing? My initial thought was to re-define the purpose of the ```-f``` value. It could mean _play this file and play this file only_ - just restarting when finished. When not given it will play the first file in the directory and keep going after that. Restarts would use the ```nowPlaying``` file. I think this would solve issues that happen in both implementations as users won't give the ```-f``` flag unless that's the only file they want to play.

TomWhitwell:
That's the way I understand the -f option: I wouldn't expect to launch with -f psycho.mp4 and have the thing run all through psycho, then on to 2001.mp4 when you launch with the -f psycho.mp4 flag. 
In other words, you'd use -f once to pick a file, but when the file is re-launched (for example after power cycling) it might have moved on to another file.

robweber:
Ok, I'll re-work what I have then to account for that. The ```nowPlaying``` file basically always trumps things as long as that video exists in the directory. ```-f``` is more of a "start here" thing but once the program moves on subsequent runs ignore it in favor of the filename in the ```nowPlaying``` file. Thanks!

robweber:
I've modified the file order based on your comments. It will now work as described: 

1. if ```nowPlaying``` exists this will load first in all cases, unless the ```nowPlaying``` file does not exist anymore
2. use ```-f``` value if given and ```nowPlaying``` is missing or invalid
3. All else fails grab a random file (0 index)

This should work for the edge cases described above where the directory option may change between runs so that non-existent files aren't loaded crashing the program. I also modified the help text to make sure it's clear that ```-f``` is ignored in the ```nowPlaying``` file exists. This should avoid confusion for users thinking it will load the given file regardless.

TomWhitwell:
Sorry, I've been really slack on this, will check these and pull them this weekend, hopefully! Thank you for all this work.

qubist:
Putting this email I sent to @TomWhitwell here so other folks involved can see it.

> Hi Tom,

> I’m building a few VSMPs based of your article and code  I’m working off your v0.2-revisions branch, and I made a few improvements, too. Now I’m working on integrating all the relevant improvements from missionfloyd and robweber into the code as well. I’m finding going through commit by commit is the best way to do this, as there’s several versions of the code in each of our forks. In addition, I did a bunch of code cleanup and commenting. I also updated the code for python3 (python2 is now depreciated), and figured out a pretty consistent way to get all the dependencies running for that. I can add that to README.md.

> I’d be happy to pass all this by y’all and then get it merged into your original repo for the project so more people can benefit from all our improvements.

> I understand you don’t have too much time to work on this, and I really respect you setting boundaries like that. Since you’ve got some visibility on the repo, though, I think it would be great to have good quality code accessible for folks building the project. I’m happy to do a lot of the heavy lifting, and just need your help with merging things in. Alternatively, I’d be happy to have collaborator access and carry some of the GitHub-repo-organization work as well.

> Pretty jazzed on this project and looking forward to bringing together your, my, missionfloyd’s, and robweber's work on the code for these little devices into a nice package with all our improvements together.

> Let me know what you think!

> All the best,
> Will Harris-Braun (qubist)

> Bonus thing: I’m gonna make an Instructable for the VSMPs! And a custom CNC’d wood frame/case.

robweber:
@qubist 

Very nice - thanks for sharing all that info. My original intent with this merge request was to lay the groundwork for adding some features. I took a look at your fork of the repo and I like the improvements you made. Might "steal" some for my own fork. 

I also did a fair amount of work on this repo as well, which started as a clone of this project but gradually got more and more elaborate on me. If you can find anything in there that makes sense to port over feel free to take a look. https://github.com/robweber/vsmp-plus

robweber:
Agreed regarding keeping things as close to the original intent as possible - that's one of the main reasons I made that a whole new project. I had to weed through lots of forks and pull bits and pieces as you've described so unifying things would be great. 

In emails with Tom in the past we had also discussed streamlining the install process for new users. Admittedly this was something I wanted to tackle once the codebase was cleaned up a little. One major drawback to the current codebase is that the program isn't setup to run as a service on the Raspberry Pi OS. It's not a hard modification to make but once done would allow for the development of an installer script that sets up all that stuff for the user. missionfloyd's branch has a lot of that work already done and this PR was meant to lay the groundwork for that modification. If you can gain some traction merging the codebase I'd be happy to tackle, or collaborate, on the install script idea.

qubist:
Oh that sounds really good. Even more of an incentive for me to merge these versions together. I'd be delighted to collaborate on getting the code running as a service and then an install script! Hopefully I can get collaborator access from @TomWhitwell on this repo to get started.

TomWhitwell:
Hi, I've added @robweber @missionfloyd @qubist to the repo as collaborators because I don't want to be a bottleneck on improvement and I don't have much time to commit to it.

robweber:
Thanks Tom - I got your invite. I appreciate you being up front with your time commitment as well. Its a testament to your original idea how much interest there is in this project.

qubist:
Thanks Tom! Much appreciated :)

robweber:
Closing as this is now redundant.

PR #4, opened 2020-10-03
Various Fixes

missionfloyd:
- Extract frames to RAM instead of SD card.
- Get screen dimensions from driver. Now only one line has to be changed to in order use with other screen sizes.
- Correctly scale anamorphic videos.
- Get video framerate from ffmpeg, instead of assuming 24Hz.
- Added contrast adjustment option.
- Added MKV support (actually, any filetype supported by ffmpeg should work)
- The `--file` argument now accepts files in any location.
- Added SIGTERM handler. Gracefully exits now,
- Better handling of log files; only create log for current video.
- Create log and video directories if missing, rather than crashing.
- Added option to automatically adjust delay to compensate for time taken to display a frame.
- Added systemd service.
  To install, copy `slowmovie.service` to `/etc/systemd/system`, then run `sudo systemctl daemon-reload`
  Start/stop/restart: `sudo systemctl [start|stop|restart] slowmovie`
  Enable/disable autorun: `sudo systemctl [enable|disable] slowmovie`
- Added config file support (optional, requires configargparse)
- Added random file option
- Added video directory option

PR #29, opened 2021-03-31
The Great Code Merge

### LINKED ISSUE #7: Multiple Video Clips
First a big THANK YOU to Tom and Bryan for putting this out.  I successfully built a VSMP and really enjoyed it.  Rather than a movie, I use mine to play video clips from my daughter's wedding.  I switched to a portrait orientation by swapping the width and height values on lines 150 and 151.  With this change, portrait video clips fill the display.

My challenge is the VSMP plays only one clip over and over, when there are two clips in the Videos folder.  Both file names are displayed at startup via line 120 ('print (movielist)') and both files have a corresponding '<progress' log file in the logs folder, so the program appears to be registering both clips.

I looked at the code on lines 188-192, which seems like it should increment currentVideo.  But my novice skills can't see the problem.  In my case, the program only plays the first of the two videos in a loop.  Any help would be appreciated!

### LINKED ISSUE #26: The same movie after each reboot
Hi!

I loaded the new version with the frame-counter and correct movie-changer. This is working perfect now - thanks a lot for this!

But now I have one more problem: I start with "python slowmovie2.py -d 120 -i 24". I thought, that every time I start without a filename (-f xxxx.mp4), an other movie will start - but every time the same movie will start and continue where it ends. How can I change this, to start a random movie after a reboot?

qubist:
### Goals
* Maintain the basic functionality of Tom's original code
* Don't add any big new features
* Polished it all up (comment, organize, refactor)
* Add a few conveniences/small features
* Set the groundwork for further expansions that others might want to make on their own
* Set the groundwork for setup as a service, and an easy installation script

Hopefully after this, the code will be more accessible to new people making the Very Slow Movie Player!

### Contributors
Code manually combined from @qubist's branch, @robweber's pull request, and @missonfloyd's pull request. Thanks to all who contributed ideas and implementations! Sorry if I picked someone else's 🙃.

#### Major
- [x] Add .mkv support
- [x] Add runtime estimates to startup debugging info
- [x] Standardize code to single quotes
- [x] Port code to python3
- [x] Standardize code to fstrings
- [x] Create logs dir if missing
- [x] Create vid dir if missing
- [x] Recognize case-insensitive file extensions
- [x] Add contrast adjustment
- [x] Add defaults section for easy configuration
- [x] Rework file-management and detection
- [x] Add optional delay compensation
- [x] Add optionally looping one video
- [x] Allow specification of video directory
- [x] Add ability to understand multiple file extensions
- [x] Add ability to play files in random order
- [x] Add file time-estimates
- [x] Refactor video info calculations
- [x] Update README to include python3 installation instructions
- [x] Add MIT License
- [x] Remove helloworld.py and replace with updated test instructions

#### Minor
- [x] General refactoring
- [x] Update print text of print statements
- [x] Update help text to match standard
- [x] Code cleanup and modernization/pythonicization

resolves #26
resolves #7
supersedes and contains #23 
supersedes and contains #4


qubist:
TODO:
* Fix the last bugs from last big commit
* Add features from robweber's PR
* Otherwise finish changelog list above

missionfloyd:
Rather than copying and pasting all my commits, it probably would've been easier to fork mine and work off of that.

robweber:
Lots going on here!

With missionfloyd's stuff merged into master already we are dumping a lot of changes at once and I agree we're likely to see some duplication. Let's give a day or two for review on this one if possible please. 

Now that we have access to the full repo, thanks Tom, we could also use the Milestone feature to plot out some of the larger upcoming tasks, like overhauling the install guide. 

Since we have more cooks in the kitchen now I'd also like to propose we add a license to this project. Something as permissive as possible but since we're taking code from multiple sources we should be upfront on what the licensing is so no one gets upset about it. Thinking MIT or similar. Thoughts?

qubist:
MIT license sounds great, but that's maybe the one thing we do need input from @TomWhitwell on, since it's lots of his original work.

Don't really know what happened with missionfloyd's recent merge. Oh just saw their update. Basically once qubist:master is done, equivalent to TomWhitwell:master, and stable, and reviewed (thanks robweber's suggestion), then we'll merge such that TomWhitwell:master matches qubist:master exactly, and then move on from there with more typical branches for new features.

TomWhitwell:
Hi All, really excited to see progress and collaboration here. 

I'm very happy to go with MIT Licence if you are all happy with that. I feel (hope) very little of the code I actually wrote will be here for very long! 

My only short term concern is that I'm able to update [the Medium post](https://debugger.medium.com/how-to-build-a-very-slow-movie-player-in-2020-c5745052e4e4)  if/when the code here has changed to break those instructions. 

Of course, medium term I can also edit the post to point to installation instructions in a readme here - which would give you all more flexibility.

README.md
```diff
@@ -2,13 +2,93 @@
...
+  -a, --adjust-delay    reduce delay by the amount of time taken to display a
+                        frame
+  -c CONTRAST, --contrast CONTRAST
+                        adjust image contrast (default: 1.0)
+```
+
+## Maintainers
+
+* @qubist
+* FIXME: @robweber and @missionfloyd you wanna be listed here?
```
robweber, line 78: My name here is good

slowmovie.py
```diff
@@ -10,205 +10,349 @@
...
+        return dir
+    # Make default videos directory if it doesn't exist and it's needed
+    elif dir == 'Videos':
+        os.mkdir(dir)
+    else:
+        raise argparse.ArgumentTypeError(f"Directory '{dir}' could not be found")
+
+# Calculate how long it'll take to play a video.
+# output value: 'd[ay]', 'h[our]', 'm[inute]', 's[econd]'
+def estimate_runtime(delay, increment, videoLengthS, videoFPS, output):
```
robweber, line 104: Could a default for the ```output``` parameter be something like a "guess" for the best output? I think this is better than hardcoding "days" as for the call below. Check if  60 > s, check min, if 60 > min, show hours, etc.

slowmovie.py
```diff
@@ -10,205 +10,349 @@
...
-        currentPosition = 0
-        log = open(logdir + '%s<progress'%currentVideo, 'w')
-        log.write(str(currentPosition))
-        log.close() 
-    
-        thisVideo = movieList.index(currentVideo)
-        if thisVideo < len(movieList)-1:
-            currentVideo = movieList[thisVideo+1]
+# ...then try the nowPlaying file, which stores the currently-playing video...
+if not currentVideo and os.path.isfile('nowPlaying'):
```
robweber, line 203: Based on the ordering of this the --file flag __always__ trumps the ```nowPlaying``` file? Is this problematic. Every time you restart the program, or the entire OS, you'll restart the video instead of loading the last played frame? Would it be better to  check ```nowPlaying``` first? Trying to think of a way to resume on subsequent runs better.

slowmovie.py
```diff
@@ -10,205 +10,349 @@
...
+    print('No videos found in video directory')
+    sys.exit()
+
+with open('nowPlaying', 'w') as file:
+    file.write(currentVideo)
+
+# Keep track of what video we're playing and which ones are available
+if not args.loop:
+    videos = list(filter(is_vid, os.listdir(viddir)))
+    fileIndex = videos.index(currentVideo)
```
robweber, line 231: Could we move this to a function and do this check when needed? The contents of this directory may change over the days/weeks so getting the index at the start may be wrong by the time we need it.

slowmovie.py
```diff
@@ -10,205 +10,349 @@
...
+
+print(f'Update interval: {args.delay}')
+if not args.random_frames:
+    print(f'Frame increment = {args.increment}')
+
+print(f'With these settings, each minute of 24fps video would take {estimate_runtime(args.delay, args.increment, 60, 24, "d")} to play.')
+print(f'A 120-min movie would last {estimate_runtime(args.delay, args.increment, 120*60, 24, "d")}.')
```
robweber, line 238: This is where "guess" as the output is better than days. What if the playback doesn't take days but rather hours? Also, why not wait for this output until we load the video info and use that instead of hardcoded fps and video length?

slowmovie.py
```diff
@@ -10,205 +10,349 @@
...
+
+# Make sure video file passed into CLI is in the videos directory
+if args.file:
+    if args.file in os.listdir(viddir):
```
robweber, line 242: Should this check be done sooner? Might be a hard stop here or at least pull in a random file or something.

README.md
qubist: Changed and re-requested :)

slowmovie.py
```diff
@@ -10,205 +10,349 @@
...
+if not args.random_frames:
+    if args.start:
+        currentFrame = clamp(args.start, 0, videoInfo['frame_count'])
+        print(f'Starting at frame {args.start}')
+    elif os.path.isfile(logfile):
+        # Log files store video progress
+        # Get the stored position for this file from its log
+        with open(logfile) as log:
+            try:
+                currentFrame = clamp(float(log.readline()), 0, videoInfo['frame_count'])
```
robweber, line 271: This may sound radical by why not refactor the log file purpose. There is a new log file for each video - to what purpose. We could store the current frame in ```nowPlaying``` as a dict ```{'file': 'video.mp4', 'frame': 29 }``` and then unify the log into a single ```slowmovie.log``` where this kind of info is displayed on each pass. Using the logging module instead of ```print()``` statements and writing files might be good too. Perhaps too big for this PR - separate PR to overhaul logging?

slowmovie.py
```diff
@@ -10,205 +10,349 @@
...
+                    # Inrcrement fileIndex or wrap around
+                    fileIndex += 1
+                    if fileIndex >= len(videos):
+                        fileIndex = 0
+
+                    # Update currently playing video to be the incremented one
+                    currentVideo = videos[fileIndex]
+
+                # Note the new video we picked in nowPlaying file
+                with open('nowPlaying', 'w') as file:
```
robweber, line 337: should this file be a global constant ```NOW_PLAYING="nowPlaying"``` and then use a function to read/write the files? Just cleaner but not necessary. One mistype on the filename somewhere would mess things up.

robweber (changes requested):
Left a lot of comments in here. Some might be better for another PR but discussion at this point would be good I think. 

Also - can we remove all the ```.DS_STORE``` files and put those in ```.gitignore```?

slowmovie.py
robweber: Looking at this more moving this to a ```get_next_video(viddir, currentVideo=None)``` type function will also eliminate some dup code lines 212 - 218. Essentially if currentVideo is None then just return the 0 indexed video.

slowmovie.py
missionfloyd: The current position is stored in per-video files in the logs folder. It has nothing to do with the --file argument or the nowPlaying file.

slowmovie.py
missionfloyd: In my fork I've removed this altogether, allowing for files from any folder.

slowmovie.py
missionfloyd: IMO, saving progress works fine the way it is. Of course, the logs folder isn't really for logging, is it? Perhaps it should be renamed.

robweber: It's more saved data, that's true. As we move more towards a service architecture moving the ```print()``` statements to an actual log file would help with debugging but not really necessary at this stage. It would help file IO (only read one file with both values instead of 2) but I agree it's not the end of the world. We're not trying to fly a rocket or anything here with this code.

slowmovie.py
qubist: I considered this but was lazy. This is the push I needed. Added in 42121a9

slowmovie.py
qubist: This is a great line of discussion, thanks for emphasizing it, Rob. The way I imagine this being used is that I start it up with something like `slowmovie.py -f 'spiritedaway.mkv'` to pick what movie to start with. But then I set the pi every reboot to run `slowmovie.py` without any file options. This way when I reboot, it'll prioritize the nowPlaying file if that's changed because I finished a movie.

Notice that's also what's instructed in [Tom's original post](https://debugger.medium.com/how-to-build-a-very-slow-movie-player-in-2020-c5745052e4e4):
```
cd SlowMoviesudo
python slowmovie.py
```

This order works with that usage.

slowmovie.py
missionfloyd: The way I have it working is that --file takes precedence over --dir. The video directory gets set to that file's directory, then plays through starting from there.

slowmovie.py
qubist: Beautiful. Implemented in db61009. I also refactored getting a random video file which appeared twice.

slowmovie.py
robweber: Nice, that's an elegant solution to that problem. That would still make the options mutually exclusive of each other though right? Passing in ```--file`` and a ```--dir``` option really just ignores the second one. From a usability standpoint forcing one or the other might make more sense for data verification. 

Probably beyond the scope of changes for this PR specifically. Could be separate PR to add that functionality later. The ordering ordering does work in the current code, just some odd edge cases.

qubist: I think it needs to be there without major changes, because we navigate to the file by appending the Videos directory path with `currentVideo` which is a filename. In missionfloyd's fork there's more complicated path stuff going on. I figure on this we can stay relatively true to the original code without compromising too much—it's not too hard to move files into the Videos directory.

(BTW if you're wondering why I don't check this in an argparser typechecking function, it's because I read in their docs that it's better to have most of that kind of logic outside of the parser.) Ok if I resolve this, while noting it for further development?

slowmovie.py
robweber: I'm going to resolve this in the interest of moving this PR along. Overhauling the logging doesn't seem like it fits with this PR. I __do__ think that a unified log file to replace the ```print()``` statements is a good idea; as is storing the current frame information in the ```nowPlaying``` file; however this isn't the time for those kind of changes.

slowmovie.py
qubist: A bit more: You're right this can cause a hard stop. Made it raise a file not found exception in 04fd8c9

slowmovie.py
robweber: Upon review maybe I'm just being picky here :) It's fine the way it is for now. If something else requires touching the file IO code in the future it can be dealt with then. There will always be more PRs. @qubist go ahead and resolve this unless you plan on doing something with it.

qubist: 5bebbc7 Implemented. Oops I named that commit wrong. Should say "...writing to nowPlaying file". Reading from it is not refactored, but it's only done in one spot so I'm leaving it

qubist:
I think I've addressed all that! 😅

slowmovie.py
qubist: I like where this is going, especially spinning it off into a separate enhancement. How do y'all feel about renaming 'log' terminology to 'progress', in preparation for that future work?

This would be renaming the log directory + the terminology in the code. Thoughts?

slowmovie.py
qubist: Whoops saw your update too late :)

qubist: @robweber we can rollback that commit if what I did is less desirable than the way it was. Such a little thing it's agonizing to decide XD

README.md
```diff
@@ -2,13 +2,96 @@
 
 ![](Extras/img.jpg)
 
-## Python / Raspberry Pi interpretation of Bryan Boyer's Very Slow Movie Player  
+[![standard-readme compliant](https://img.shields.io/badge/readme%20style-standard-brightgreen.svg?style=flat-square)](https://github.com/RichardLitt/standard-readme)
 
+> Very Slow Movie Player using Python + Raspberry Pi
```
robweber, line 7: I was reading the guide for the Standard README format (nice find by the way, haven't used that before) and saw this: https://github.com/RichardLitt/standard-readme/blob/master/spec.md#short-description 

Should the ```>``` be removed here if this is technically the "short description"? See bullet 3 in that link.

helloworld.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
```
robweber, line 1: Just a comment on this file in general - do we even need it? Running the main program with no arguments ```python3 slowmovie.py``` is basically the same thing. It will load the test move from the default Videos directory and play it. I had issues getting this to run when cloning this PR and it got me wondering why duplicate all this code for a test case?

robweber (commented):
I ran this branch on a few different arg options and then a longer test of about 30 min. No weird errors or things that I could find.

README.md
qubist: It looks like [in the past](https://github.com/RichardLitt/standard-readme/commit/2986c499017e94aa991e93180e717f5a80e44ca2#diff-bc6661da34ecae62fbe724bb93fd69b91a7f81143f2683a81163231de7e3b545R40) it's required that it starts with `>` and [they changed it](https://github.com/RichardLitt/standard-readme/commit/0211114475754c064f643b6e36b382e9c9298312#diff-bc6661da34ecae62fbe724bb93fd69b91a7f81143f2683a81163231de7e3b545R81). Fixed in 06b1963, good catch, Rob

helloworld.py
qubist: That's a good point, maybe it's a relic of the original development process... Including the test.mp4 file kind of fulfills the same purpose

LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 FIXME
```
qubist, line 3: We should figure out who to put here

slowmovie.py
```diff
@@ -10,205 +10,372 @@
...
+from PIL import Image, ImageEnhance
+
+# Ensure this is the correct import for your particular screen
+from waveshare_epd import epd7in5_V2 as epd_driver
+
+def clamp(n, smallest, largest):
+    return max(smallest, min(n, largest))
+
+# Handle when the program is killed and exit gracefully
+def signal_handler(signum, frame):
```
qubist, line 23: I noticed [this commit](https://github.com/tkameroski/SlowMovie/commit/1efaec0c5b8e9b2ee60382960c11c27d3bff5622) from @tkameroski. What's the difference between these approaches? Why does one work with ctrl-C? Thoughts anyone?

README.md
```diff
@@ -2,13 +2,96 @@
...
 
+In August 2020, Tom Whitwell posted ["How to Build a Very Slow Movie Player for £120 in 2020"](https://debugger.medium.com/how-to-build-a-very-slow-movie-player-in-2020-c5745052e4e4), with instructions on how to build a VSMP with the new [7.5-inch, Raspberry Pi-compatible e-paper display from Waveshare](https://www.waveshare.com/product/displays/e-paper/epaper-1/7.5inch-e-paper-hat.htm).
 
+SlowMovie is the code that runs a VSMP on your Raspberry Pi.
+
+## Install
+
+**Note:** These installation instructions assume you have access to your Raspberry Pi and that you have the hardware set up properly. See the [Medium post](https://debugger.medium.com/how-to-build-a-very-slow-movie-player-in-2020-c5745052e4e4) for more complete instructions.
+
+SlowMovie requires [python3](https://www.python.org/). It uses [ffmpeg](https://ffmpeg.org/) via [ffmpeg-python](https://pypi.org/project/ffmpeg-python/) for video processing, and [PIL](https://python-pillow.org/) [FIXME is this the right link?] for image processing.
```
qubist, line 21: Any ideas whether https://python-pillow.org/ is the right link for PIL that we're using?

helloworld.py
qubist: If it feels like we should take action on this, would we delete the helloworld file and change instructions on [line 39 of README.md](https://github.com/TomWhitwell/SlowMovie/pull/29/files/06b1963d2c2a62dd16195252e6a886b7ce67f42e#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R39)?

slowmovie.py
robweber: Comparing the PR to this commit the only difference is the removal of ```epd_driver.epdconfig.module_init()``` right? 

The approach in this PR is what I'd call the _correct_ one based on how the waveshare library is constructed. If you don't call init() before the exit() function you get RunTime error when hitting CTL-C. The epd driver calls ```SPI.close()``` when you use the sleep() function. Because of this the exit handler is calling close() on an already closed object. You have to init first, and then close it properly. It's bit odd since you want all the other clean-up operations that exist in the ```module_exit()``` function to happen but the SPI call will thrown an error if you don't open it first. 

I tested this again with the PR code by removing the init() call and you get a RunTime error. Putting it back in you get a clean exist.

README.md
robweber: I believe that is the right library for Python 3. Way more info than I cared to know put apparently "PIL" was the python 2 version that has been abandoned. Pillow is the currently active Python3 version but you still use "import PIL" to maintain backward compatibility. Guess I learned something today.

LICENSE
robweber: I believe it should be Tom as the owner of the original project. My understanding of it is that the original author license's the work. If you modify the code you retain your right to your changes but when you push back to the original repo you give the author rights to distribute your changes under the original license. I'm not a lawyer so but that's how I always took it. I imagine in a real dispute you'd argue over if you implicitly gave over your permission with the PR or if an additional agreement was needed but this is a pretty small project with a very permissive license so hopefully we're all cool. 

I checked a few other projects and most people don't even bother to change those lines which was kind of funny.

helloworld.py
robweber: That was my thought, change the test instructions to just try running the program without args and load that test file. Just looking at it now it has a lot of pieces that were modified in the main program, it's already quite out of date and will only get worse.

slowmovie.py
missionfloyd: Using try works too
```
def exithandler(signum, frame):
    try:
        epd_driver.epdconfig.module_exit()
    finally:
        sys.exit()
```

README.md
qubist: Hehe thanks for doing that research! Resolved in 98c1801

slowmovie.py
qubist: I think the way signal_handler is currently defined (from @robweber's PR) matches @tkameroski's approach. And, as opposed to @missionfloyd's try: finally: approach, it calls `module(exit)` (plus the accompanying clean-up operations) every time instead of only when SPI is open. If I understand all that correctly. So I think this can be resolved with no changes.

LICENSE
qubist: @TomWhitwell are you cool with that?

helloworld.py
qubist: Implemented in 218a84f. We'll have to make sure Tom's medium post update includes this as it causes a big change in the instructions.

slowmovie.py
missionfloyd: My way does call module_exit() every time. It'll obviously fail if SPI isn't open, thus the try statement. Anyway, that commit message is incorrect; it already worked with ctrl-c.
```
signal.signal(signal.SIGTERM, exithandler) # 'kill' / service stop
signal.signal(signal.SIGINT, exithandler) # Ctrl-C
```
What it really did is make it work with waveshare's recent updates, which close SPI every time sleep() is called. Same thing https://github.com/missionfloyd/SlowMovie/commit/01ff5cf3c0ff294f9d2842be27d5c0249b99bddf does.

LICENSE
TomWhitwell: Yes I'm cool with either no name, my name or lots of names from contributors.

helloworld.py
TomWhitwell: The idea for helloworld.py was to have a very minimal program that users who wanted to learn or develop could look at to get a sense how to call the epaper, extract images from file, dither those images, without all the overhead of config etc.
Maybe just move helloworld to the extras folder?

slowmovie.py
robweber: @missionfloyd  I don't think anyone is arguing that your version doesn't fail gracefully, it does. The difference is that if the SPI.close() method fails and then throws an error you miss any other processing that was done in the [module_exit() ](https://github.com/qubist/SlowMovie/blob/master/e-paper/rpi/python/lib/waveshare_epd/epdconfig.py#L81)function. By using ```module_init()``` first the ```module_exit()``` function gracefully exits and the rest of the operations complete. 

I totally agree the waveshare code should be better about this and honestly it's on them to make these kind of checks so methods don't fail but they don't.

robweber (approved):
I think I'm good!

helloworld.py
qubist: Sounds good, Tom. I'll add it back in and edit it to match `slowmovie.py` a bit more.

LICENSE
qubist: Resolved by 6989241

helloworld.py
qubist: Thanks for the input, @TomWhitwell, I see now that it's important to have a simpler case that people can learn from. I added it back in the Extras folder via e1e3635

missionfloyd:
I've merged my stuff to a new branch, in case anyone feels like doing this the right way.
https://github.com/TomWhitwell/SlowMovie/tree/new

qubist:
Closing this to incorporate missionfloyd's changes properly

slowmovie.py
missionfloyd: It doesn't need to do the rest of it because it'll have already been done. That's the whole reason it fails.

robweber: I don't see any other place the GPIO gets closed properly prior to this. The SPI call happens first and that's where the error will happen.

missionfloyd: After every frame, [epd.sleep()](https://github.com/qubist/SlowMovie/blob/e1e36358cc78cfa9bd40f4231cdf46c8b2337c11/slowmovie.py#L376) is called, which in turn calls [epdconfig.module_exit()](https://github.com/waveshare/e-Paper/blob/af9694c602fb24d776cd0f0b3e073a477294e3d8/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd7in5b_V2.py#L170-L178). If you stop it while it's sending a frame, SPI is still open, so this doesn't fail.

robweber: You're linking to code in two different repos. That's probably why our thoughts on this differ. The waveshare code, as it exists here in qubist:master, doesn't call module_exit() when you use sleep. It calls [module_sleep() ](https://github.com/qubist/SlowMovie/blob/e1e36358cc78cfa9bd40f4231cdf46c8b2337c11/e-paper/rpi/python/lib/waveshare_epd/epd7in5_V2.py#L168)which only closes SPI and that's it. The current version of this program bundles the waveshare lib with the ```slowmovie.py``` file. Probably not best practice but we can't make decisions in that code based on code we're not bundling with the it. 

We should probably test the main waveshare stuff again now that it seems to have incorporated fixes and think about unbundling it from this repo.

missionfloyd: Yeah, the waveshare driver here is outdated. It should probably be removed.

TomWhitwell:
Hi, I just added this line to the Medium post: 

image

Issue #31, opened 2021-03-31
Add Unified Logging


robweber:
Replace the ```print()``` statements with a unified log file using the Python logging module. The waveshare library already has support for that so we'd capture some of that output too. 

When the program is run as a service a log file has more usefulness than the command line output for monitoring and debugging. 

_Originally posted by @robweber in https://github.com/TomWhitwell/SlowMovie/pull/29#discussion_r605086464_

missionfloyd:
If you run it as a service, all output goes into the journal, viewable with `journalctl -u slowmovie`. This is also why I removed some of the print statements, thousands of lines of "Displaying frame x of y"

robweber:
That's a good point, which may solve this problem. There may be value in a debug flag and then pass all print statements through a function. That way you could debug on the command line but when running normally you wouldn't spam the log. Definitely want some type of solution as this moves more towards a service orientation as opposed to all the command line setup.

missionfloyd:
Another option could be to do it the other way around. Add a [hidden](https://docs.python.org/3/library/argparse.html#help) `--service` argument to suppresses excess output.`

missionfloyd:
I added the `--service` flag to suppress excess messages. The service file will need to be copied again.

missionfloyd:
Okay, here's how I see it. There are three places output can end up: console, systemd journal, and a log file. When running as a service everything printed to the console automatically goes to the journal. When not, it should go to a log file. The logging module adds logging level and such, but that can all be customized. That information probably shouldn't go to the console, but it could go into the log file. Or not. Maybe just timestamps.

Anyway, just my ramblings on the matter.

PR #36, opened 2021-04-02
Add missionfloyd changes

qubist:
Ok, let's add missionfloyd's changes and then add any edits or additional changes from the `code-merge` branch on top of that. I still think we should hop on a call and talk synchronously about this repo and how to move forward, though!


robweber:
This is getting very confusing. The former PR was like 98% complete. I'll try and take some time to review these later. I also see a branch just called "new" - not sure what that is all about. 

I highly suggest going forward that everyone take a deep breath - things can always be merged in later and your specific way of tackling an issue isn't always the best way. When you have more than one person proposing changes it's a compromise sometimes, not the same as when you hack away on your own projects. There is a wider audience to think about and based on the current Issues there are some very basic changes that can be made to make things easier for people.

qubist:
The forer PR was based on v0.2-revisions which was based on these above commits.

I totally agree with you @robweber  that collaboration stems from the conversations we have while implementing features and it's a different mode than when hacking on an individual project. The old 98% PR was kinda 1/4 my personal hacking changes, and then the next 3/4 collaboration.

I think we should merge the old 98% PR and move on, but also want to respect @missionfloyd. Maybe it's just time to stick to that PR and move ahead, though.

missionfloyd:
Maybe I've been working on this by myself for too long, but it felt like my PR was being clobbered. So thank you.

qubist:
I can totally see how it felt like that. I think I would've felt the same way in your position.

qubist:
Would it work for you, @missionfloyd, if we actually do the following instead of merging this branch?

1. Merge the `v0.2-revisions` branch into `master`.
   * (In commit e64538c in that branch, Tom added a bunch of your changes from above)
   * (That's also the branch that the fork qubist:SlowMovie was based on, which then lead to the "Great code merge")
2. Merge the `code-merge` branch into `master`.
   * This is extra changes that were built on top of `v0.2-revisions`, plus some of your changes reinterpreted, plus collaborations between all of us on the Readme and other things
3. Continue in the future to add the features from your `new` branch (same as your PR) to `master` in a collaborative fashion
   * (For this we would also use a PR-per-feature approach and request reviews from each other)

I'm hoping that achieves the following goals:
* doesn't clobber your PR and includes your contributions (via Tom) to `v0.2-revisions`
* Follows (ish) the overall progression of the codebase over time
* Includes the more collaborative work done in #29 
* Sets up future development for success

Let me know what you think.

missionfloyd:
That sounds like that's right back to where it was before.

qubist:
Would you mind clarifying what you'd like to see instead? I think that solution is pretty fair to both your contributions and the work that's been done since.

missionfloyd:
Ideally, the PR's that were already there would've been merged, and then go off of that. The code in this branch has been running for several months on a raspberry pi zero. It works well. 

Basically, you'd bring your changes over here, instead of taking over mine. I think it would be simpler than manually frankensteining everything together.

robweber:
Just trying to make some sense of the different PRs and branches is this correct? 

1. [new](https://github.com/TomWhitwell/SlowMovie/tree/new) - this branch is the proposed merge happening here which is just a mirror of the current missionfloyd master branch right? 
2. code-merge - has the changes from #29 that is in the [Community Changes](https://github.com/TomWhitwell/SlowMovie/pull/37) PR
3. [v0.2-revisions](https://github.com/TomWhitwell/SlowMovie/tree/v0.2-revisions) - we're kind of abandoning this as these changes are covered in one of the other branches (at least there is no PR for it)

I don't want to take the time to review all of these if that's not the flow. On a related note, I had made a tag for the master branch a day or two ago. I wanted to make sure we had a snapshot of all Tom's original work before merging anything into master. It doesn't look like that's there any more - any ideas what happened?

missionfloyd:
That's about right.

After I reverted my overzealous merge the other day, I rolled master back to how it was before I touched it so that the file histories weren't overrun with "Revert commit." That might be why the tag is gone

helloworld.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
robweber, line 1: per the [conversation with Tom](https://github.com/TomWhitwell/SlowMovie/pull/29#discussion_r606103910) the idea of this file is to be kind of a minimal "proof of concept" for people to see how the system worked. If that's the case maybe we could strip this __way__ down. Basically hardcode the test file and pull random frames from it. Less code to maintain as changes happen going forward. Could do this in a new PR after the merge.

.gitignore
```diff
@@ -1,19 +0,0 @@
-# Thumbnails
```
robweber, line 1: Can we add the .gitignore back in? Nice to make sure we keep system files from various dev environments out of the repo. The .DS_Store files for example.

slowmovie.py
```diff
@@ -10,205 +10,233 @@
...
-from PIL import Image
+import os, time, sys, random, signal
+from PIL import Image, ImageEnhance
 import ffmpeg
-import argparse
+from fractions import Fraction
 
-# Ensure this is the correct import for your particular screen 
-from waveshare_epd import epd7in5_V2
+try:
```
robweber, line 18: Is it worth a try/except here? I think the configargparse library should either be a lib requirement or not. If we have good instructions on what modules to install it should be available. I understand the "fail gracefully" idea here since argparse shares the same method calls. I think it's confusing though to people who may try to use a conf file and it just doesn't work and they can't figure out why.

slowmovie.py
```diff
@@ -10,205 +10,233 @@
...
+    if os.path.isdir(value):
+        return value
+    else:
+        raise argparse.ArgumentTypeError("Directory '%s' does not exist" % value)
+
+def supported_filetype(file):
+    _, ext = os.path.splitext(file)
+    return ext.lower() in fileTypes
+
+def video_info(file):
```
robweber, line 74: @qubist had some great code here to catch issues with loading MKV files - guessing we'll merge that in after this as part of Community Changes PR?

slowmovie.py
```diff
@@ -10,205 +10,233 @@
...
+
+os.chdir(os.path.dirname(os.path.realpath(__file__)))
+
+if "configargparse" in sys.modules:
+    parser = argparse.ArgumentParser(default_config_files=["slowmovie.conf"])
+else:
+    parser = argparse.ArgumentParser()
+parser.add_argument("-D", "--dir", default = "Videos", type = check_dir, help = "Select video directory")
+parser.add_argument("-r", "--random", action = "store_true", help = "Display random frames")
+parser.add_argument("-R", "--random-file", action = "store_true", help = "Play files in random order")
```
robweber, line 90: aren't ```-R``` and ```-f``` kind of mutually exclusive? Should we put them in an argparse group to reflect that?

slowmovie.py
```diff
@@ -10,205 +10,233 @@
...
-    epd.init()
+    # Open frame.bmp in PIL
+    pil_im = Image.open("/dev/shm/frame.bmp")
 
+    if args.contrast != 1:
+        enhancer = ImageEnhance.Contrast(pil_im)
+        pil_im = enhancer.enhance(args.contrast)
 
+    # Dither the image into a 1 bit bitmap
+    #pil_im = pil_im.convert(mode = "1", dither = Image.FLOYDSTEINBERG)
```
robweber, line 202: This exists in other versions of SlowMovie - is it not needed?

slowmovie.py
```diff
@@ -10,205 +10,233 @@
...
 
+    # display the image
+    #print(f"Displaying frame {currentFrame} of '{videoFilename}'")
```
robweber, line 205: Can we add this back in? I know it spams the log but we can clean up logging in #31

slowmovie.py
```diff
@@ -10,205 +10,233 @@
...
+    parser = argparse.ArgumentParser()
+parser.add_argument("-D", "--dir", default = "Videos", type = check_dir, help = "Select video directory")
+parser.add_argument("-r", "--random", action = "store_true", help = "Display random frames")
+parser.add_argument("-R", "--random-file", action = "store_true", help = "Play files in random order")
+parser.add_argument("-f", "--file", type = check_mp4, help = "Specify an MP4 file to play")
+parser.add_argument("-d", "--delay", default = timeInterval, type = int, help = "Time between updates, in seconds")
+parser.add_argument("-i", "--increment", default = frameIncrement, type = int, help = "Number of frames to advance on update")
+parser.add_argument("-s", "--start", type = int, help = "Start at a specific frame")
+parser.add_argument("-c", "--contrast", default=contrast, type=float, help = "Adjust image contrast (default: 1.0)")
+parser.add_argument("-a", "--adjust-delay", action = "store_true", help = "Reduce delay by the amount of time taken to display a frame.")
```
robweber, line 96: Why is adjust delay optional? Wouldn't you always want the delay to be as accurate as possible? One less command line argument.

robweber (commented):
Other than what's missing from Community Changes looking pretty good. @missionfloyd said it's been running solid.

I think we should also append the updated README and License file from the other PR though into this one. We'll want updated instructions if we merge into master. This branch removes the waveshare lib which is kind of a big deal. We'll want instructions for people to clone and build that library separate.

helloworld.py
missionfloyd: This file currently does something slowmovie.py doesn't: play random frames from the whole folder. That can probably be added, and this one stripped down to serve as example code.

slowmovie.py
missionfloyd: Yeah, that'll go back. I had no idea MKV wasn't working right, I thought I'd tested it.

slowmovie.py
missionfloyd: The driver already does this.https://github.com/TomWhitwell/SlowMovie/blob/e66293e204ef933f284a5d8c8cdcee7fb99e0594/e-paper/RaspberryPi%26JetsonNano/python/lib/waveshare_epd/epd7in5_V2.py#L118
(Floyd-steinberg is the [default](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.convert), so that parameter isn't necessary)

slowmovie.py
missionfloyd: Sure, go for it. 

(I don't know is you know this, but you can add code suggestions to your comments with the ± button. Then they can be  committed right from here.)

slowmovie.py
missionfloyd: I actually don't know why you wouldn't want it. I put that as a 'just-in-case', but I suppose it isn't really needed.

.gitignore
missionfloyd: That should go back in, yes.

slowmovie.py
missionfloyd: I was originally trying not to add more dependencies, but I've been working out saving nowPlaying to the config file instead, which would make it a requirement anyway, so might as well.

slowmovie.py
missionfloyd: You're right, they are mutually exclusive (-f overrides -R.) That's a good idea.

missionfloyd: I added it for now, but I'm not sure it's worth it. It might be simpler to explain the order of precedence. Or maybe there's a better way to do all this. Your idea for playback modes could work.

See, what I thought might be a good idea was to write the current settings to the config file. That way the script can be started again with no arguments and continue with the previous settings. But then, the only way to unset some things would be to edit the config file. I might have to ditch that idea if it makes things too complicated.

robweber: Is there any reason why ```-f``` and ```-D``` need to be mutually exclusive? ```-D``` sets the directory and ```-f``` sets the file to play from that directory. Why shouldn't a user be able to do both of those things? I see in the code, further down, where the directory is automatically set to the directory containing ```-f```  anyway. 

I think your check works to at least make the order of operations somewhat explainable. Everything won't be perfect in this one PR so as long as it's explainable the entire file loading order can be tackled in another run at it.

qubist (commented):
This looks great! Thanks for all this work.

qubist:
> I think we should also append the updated README and License file from the other PR though into this one. We'll want updated instructions if we merge into master.

I can go ahead and do this if that's cool with y'all.

> This branch removes the waveshare lib which is kind of a big deal. We'll want instructions for people to clone and build that library separate.

Can someone clarify for me why this was removed? (Oh, is it because [missionfloyd's improvements were merged into their code](https://github.com/waveshare/e-Paper/commit/af9694c602fb24d776cd0f0b3e073a477294e3d8) so we can use that now?) And also what's the new way to install it? I feel happy to update the instructions in README.md once I understand these motivations.

robweber (approved):

robweber:
It was removed as part of missionfloyd's branch changes but I can speculate that it was mostly due to best practice. Not having to drag the weight of the waveshare lib means less code duplication and we're not constantly having to merge the latest changes from their repo back into this one either. It also opens the door in the future for supporting other display types with a more generic interfacing too. If we did encountered an issue - like with the ```sleep()``` code - where waveshare needs patching we can always fork that repo and offer a fixes branch as a short term fix. 

The build process is roughly the same, but you'll need to clone the waveshare lib prior to running the ```setup.py install``` command.

qubist:
Thanks for the explanation. I'll modify the README instructions and add that file. After that file's added in, I'm good to go for merge! NB @TomWhitwell that'll be the time when the Medium post should be updated.

slowmovie.py
missionfloyd: `-f` sets the directory anyway. If a directory is passed with `-D`, and a file in a different directory with `-f`, `-D` is ignored. If `-f` is a file in the same directory, it's redundant. Could make it single argument that can take either.

I was thinking though, if this is gonna be a service, would it make more sense to have only a config file? Then changing settings would be: edit config file, then restart service (`sudo systemctl restart slowmovie`)

qubist:
Sweeeet, let's merge! Everyone else ready? @missionfloyd you wanna do the honors?

PR #37, opened 2021-04-02
Community changes

qubist:
Duplicate of #29, but on a branch instead of @qubist's fork.

Adding the changes that the four of us talked through in #29. This will include the updated README, some simplifications of missionfloyd's version to better match Tom's original code, and probably other tweaks to find the best parts of both branches.

Would be merged after #36


robweber:
I guess with the PR changes I missed why helloworld was added back in. I gather a discussion was had but just curious.

qubist:
That makes sense, sorry about that. Here's a [link to the conversation.](https://github.com/TomWhitwell/SlowMovie/pull/29#discussion_r606103910)

qubist:
Just to be clear this is basically a duplicate of #29 but on a branch of TomWhitwell's repo instead of qubist's. I know it's confusing but I think this way we acknowledge everyone's contributions while still getting the desired result that we collaborated on in #29. With only a bit extra work..

README.md
```diff
@@ -2,13 +2,96 @@
...
+  * `git clone https://github.com/TomWhitwell/SlowMovie/`
+  * Navigate to the new SlowMovie directory: `cd SlowMovie`
+2. Install e-paper drivers
+  * Go into the e-paper driver directory: `cd e-paper/rpi/python/`
+  * Run `sudo python3 setup.py install`
+  * Go back to /SlowMovie: `cd ../../..`
+3. Make sure dependencies are installed
+  * `sudo apt install ffmpeg`
+  * `pip3 install ffmpeg-python`
+  * `pip3 install pillow`
```
missionfloyd, line 37: This isn't needed anymore, I've added it and the rest of the dependencies (RPi.GPIO/Jetson.GPIO, spidev, numpy) to waveshare's setup.py so it's automatically installed.

README.md
```diff
@@ -2,13 +2,96 @@
...
+On the Raspberry Pi:
+0. Make sure SPI is turned on
+  * run `sudo raspi-config`
+  * Navigate to Interfacing Options > SPI
+1. Clone this repo
+  * `git clone https://github.com/TomWhitwell/SlowMovie/`
+  * Navigate to the new SlowMovie directory: `cd SlowMovie`
+2. Install e-paper drivers
+  * Go into the e-paper driver directory: `cd e-paper/rpi/python/`
+  * Run `sudo python3 setup.py install`
```
missionfloyd, line 32: Pip will need to be installed before setup.py can be installed (sudo apt install python3-pip)
After that, setuptools may need to be updated (sudo pip3 install setuptools -U)

README.md
robweber: Here's a [link to that](https://github.com/waveshare/e-Paper/blob/master/RaspberryPi_JetsonNano/python/setup.py#L4) in the waveshare driver just to confirm. 

I'm of mixed opinion on this one as we are importing PIL as part of ```slowmovie.py``` - its not truly an external dependency that only the waveshare driver uses. If all the PIL code was contained with the EPD driver I'd say that it's that package's responsibility but we're using it too. I guess to put it another way imagine I had program `foo.py` that used library `bar`. `bar` uses `fancylib` as a dependency. That's awesome because I want to use `fancylib.do()` as part of my code too so I import `fancylib` and use it all I want. Now `bar` is an upstream library maintained by other people - they move on and no longer want to use `fancylib` any more. Now my code is broke because `fancylib` is no longer installed by the library and everyone using `foo.py` is angry it doesn't work. 

Sorry for the long hypothetical. Installing it with this one line statement doesn't cost us anything but in a future where maybe a different EPD driver is being used we'll be glad it's there since we're stilling using PIL  as part of the ```slowmovie.py``` file instead of accidentally breaking things and having to remember to add it back in.

README.md
qubist: Thanks! I'm gonna add this into #36 since that's the most up-to-date version of the readme right now.

qubist: 26c3dcf

qubist:
I propose the following:

1. Close this PR
2. Merge #36 
3. I kick off the new proper one-PR-per-feature flow by creating a few new atomic PRs for the few things that were actually different about this branch and `new`:
    * Runtime estimates (0b63bbe, 42121a9)
    * .mkv support (f450a907, baa0526, 8f0c58a)
    * Standardize to fstrings (a885c8c, 5f00071)
    * Refactor code that looks at the videos directory and gets next video  (db61009)
    * Clean up help text to match format of most bash commands (2891635)
    * Anything else @robweber and I worked on in #29 that I forgot about.

Once #36 is merged, this `code-merge` branch will be based on an outdated version of `master`, which makes for messy merging. This plan resolves that (albeit at the cost of extra work by me, which I'm ok with). It also starts us down the road of good PR and review practices, where each PR is one coherent idea, and it's easy to have comments & discussion on them.

Sound good?

README.md
missionfloyd: Okay, I get that. Since we're using it independently of EPD, we should install it ourselves.

README.md
missionfloyd: Now that I think about it, git isn't installed by default either. I think on NOOBS it is, but it isn't on vanilla Raspbian.

qubist: Ooh good call. Would you mind adding bullet-points for those in #36, since you seem to have a really good handle on these dependencies? Maybe even a new step 1. called "Set up environment/dependencies" or something would be in order here. Something like:
```
0. Make sure SPI is turned on
    * ...
1. Set up environment
    * apt update you mentioned
    * git
    * Make sure pip3 is installed: `sudo apt install python3-pip`
    * Make sure setuptools is updated: `sudo pip3 install setuptools -U`
2. Install e-paper drivers
    * Clone the Waveshare repo: `git clone https://github.com/waveshare/e-Paper`
    * Go into the e-paper driver directory: `cd e-Paper/RaspberryPi_JetsonNano/python/`
    * Install the drivers: `sudo python3 setup.py install`
    * Go back out of the e-paper directory: `cd ../../..`
3. Clone this repo
    * ...
4. Make sure dependencies are installed
    * ...
5. Test it out
    * ...
```

robweber:
I think that works well. There are many minor things that we can more easily fix through PRs later. I was keen to start working on a better install/update system which relied heavily on the service pieces added in #36. I'll close out my original PR as well since that's all duplicate work at this point.

README.md
missionfloyd: I merged that branch just before reading this comment, It'll have to go here.

missionfloyd:
Sounds good to me.

README.md
qubist: Thanks for typing it all out, love the improvements you made. I saw you carried them over to a commit on `master`. Sorry about the confusion there!

PR #40, opened 2021-04-06
Automatic PEP8 linting

robweber:
More housekeeping type stuff but I think this will help maintain code consistency. This PR adds a GitHub Action to automatically run [Flake8](https://flake8.pycqa.org/en/latest/) to check for [PEP8](https://www.python.org/dev/peps/pep-0008/) code styling. This triggers on any pushes or PRs to the repo. It does exclude the `Extras` directory but that's easily brought back in if we wanted to by changing the config. 

I went ahead and ran flake8 against ```helloworld.py``` and `slowmovie.py` so that all the checks pass. No code was changed, mostly just had to add or remove spacing. As we revise this code more having an automatic check will keep things looking the same no matter who the contributor was. Just another automated check as part of the review process prior to PRs getting merged in. Might seem like overkill but even on small projects I've had this really help when you revisit the code weeks or months later to have good styling.


qubist (approved):
Looks good to me!

slowmovie.py
```diff
@@ -50,23 +58,27 @@ def generate_frame(in_filename, out_filename, time):
...
     return value
 
+
 def check_dir(value):
     if os.path.isdir(value):
         return value
     else:
         raise configargparse.ArgumentTypeError("Directory '%s' does not exist" % value)
 
+
```
missionfloyd, line 76: What's up with all these extra blank lines?

robweber: I'm guessing you mean between the function defintions? PEP8 specification is 2 spaces between function and class definitions.

robweber:
I had to update the syntax after the merge from master to make it pass the flake8 tests. I also removed the "long lines" check. We can add it back in but probably does more harm than good.

slowmovie.py
missionfloyd: Oh, okay.

missionfloyd (approved):

qubist:
Is there a good way to automatically have my code follow the linting standard (I'm using Atom, but in general)?

robweber:
I did a quick check and it looks like Atom does have some PEP8 plugins. Not sure if they auto format or just markup? I use Atom as well and have it tied to my RPI for dev purposes. I usually just run the flake8 command manually prior to pushing commits to cleanup any issues. Over time you kind of just start writing it that way by habit :)

qubist:
Gotcha. Let's put info & instructions about using flake8 in [the Contributing document](/CONTRIBUTING.md) to make it easy for folks to pass the linter.

From what I found it's:

* On your dev machine, run `pip3 install flake8`
* Before committing, lint your code by running `flake8 --ignore E501 slowmovie.py`, and fix any errors

Something like that?

robweber:
That's pretty good. I can update as part of #46. I'll put some links to the PEP8 documentation as well so everyone knows we're not just making up a system arbitrarily.

PR #41, opened 2021-04-06
Dynamic Driver Loading

### LINKED ISSUE #5: Parameterize the display and size
If you parameterize the EPD board and display size (for example, in my case I'm using an 880x528 sized 7x5 HD epd), it would be easier to simply specify a board/size instead of rooting through code to find where the EPD is defined and used.  Granted, this does not work on 'b' variants because of differences in API, but this can also be filtered and alerted upon if someone tries to use a multi-color display.

robweber:
Looking for a little help testing this one. If all works as I think it does this should close #5  as well as open the door for future devices being very easily added to the core code. The basics are that individual device displays are now configurable as an arg at runtime, with the default being the 7.5in waveshare from the README (as was always the case). 

For waveshare devices that are in the e-Paper repo these are essentially all now active be default. Just pass in the name of the display. For other types of devices the code is now extendable so they can be easily added without having to modify `slowmovie.py` for each one. 

This code is like 90% ready in my opinion but I think it could probably be optimized better. I'm super comfortable with reflection and dynamic class loading in Java but not as much in Python. This all works but could be a better way to do it. 

### VirtualDisplayDevice

Within the `displays` directory there is now a class `VirtualDisplayDevice` which all displays will extend. The child classes can control a single display, or a family of displays, as long as they all use the same underlying code. Each child has to implement, at a minimum: 

* name of the package it's implementing (ie waveshare_epd)
* a method to return a list of compatible devices
* a method to allow for updating of the display
* width/height of the display

There are additional methods for sleep, close, etc that may or may not apply to individual devices. If you need to override you can, but if you don't it won't crash anything. 

### Device Factory

The device factory class is responsible for two things: 

1. Loading a list of all compatible devices
2. Translating the command line argument from `slowmovie.py` into a concrete class that can be used

It does this by detecting all subclasses of `VirtualDisplayDevice` and getting list of their allowed devices. When a matching device is loaded it creates the child class and passes it back to `slowmovie.py`. This allows for a defined interface that `slowmovie.py` can use no matter what the device. 

### Mock Display

As a test, and reference implementation, there is a mock display (`mock_display.py`). This emulates a display that does nothing but print statements about what it is doing. I added comments here to illustrate how you'd use this as an example to define an actual working display class. 

As an added bonus you can use this display class to test code without an actual display. You'll see the print statements about what it is doing but don't actually need a display to do anything. 

### SlowMovie.py

In the `slowmovie.py` file there are a handful of changes to make this work. Instead of defining the driver specifically the `displayfactory` is loaded. The arg `--epd` can be used to pass in the display type you want to load. Once loaded the following calls will happen: 

`epd.prepare()` - called at the top of each update loop. For waveshare devices this would be the `init()` method. 
`epd.display()` - called with the image object to update the display
`epd.sleep()` - will call sleep, if that device implements it

Additionally `epd.close()` is called when the program exits. 

## Testing

You should be able to clone this branch and test. Just running `python slowmovie.py` will load the default driver with the new code. Running `python slowmovie.py -h` will show a list of compatible devices and using `-e mock_display.test` will run with the mock display enabled. 

## Help

I only have one device to test this on, the default, so any other tests that can be done would be great. Switching between the mock device and my actual device worked. I think I could easily code an implementation for the Inky devices but without something to test (or someone willing to try my code) I don't want to just blindly do it and call it done.


robweber:
Reading through some more Issues I see Inky is also a popular option (I'm sure there are more). I even found at least one fork where someone adapted the code to work with one. How crazy do we want to get here? I could adapt this to work with future types fairly easily. Here is some psuedo code off the top of my head - assume the magic `factory` class does the importing and driver loading. 

```
# define a virtual device class with a few common methods
class VirtualDisplayDevice:

  def __init__(self, deviceObj):
    self.device = deviceObj

    # helpers so we can grab width/height as part of slowmovie
    self.width = self.device.width
    self.height = self.device.height

  def display(self, image, kwargs*):
    # whatever to set the image and show it

  def sleep(self):
     # etc, etc,


# extend this for each display type - 
class WaveshareDisplay(VirtualDisplayDevice):

  def display(self, image, kwargs*):
    # display code for waveshare, other devices would put their code here
    self.device.display(epd.getbuffer(pil_im))
```

Then in the `slowmovie.py` file you'd let the user pass in the module using a predefined list. Something like `waveshare.epd7in5_V2` or `inky.impression` to load the driver. We'd have to keep a list of driver types. The factory code would encapsulate the dynamic importing and then creating the device instance. 

```
# in slowmovie.py

# load the display or throw an error if not found
display = factory.load_display(args.epd)

display.display(image)

display.sleep()

```
This would let others pretty easily adapt to new devices and get some PRs going for other types with a framework. It would be some overkill now but a bigger payoff later - less forks for specific devices. I can keep plugging away at this - thoughts?

qubist:
Looks really cool. One thing to consider is how does it change the ability for a noob to just follow Tom's directions and make a VSMP with the 7.5-in Waveshare? As long as someone who wants to do that doesn't have to interact with the idea of multiple screen compatibility, I approve.

robweber:
Agreed. I think we keep the 7.5in Waveshare as the default for all options. You should always be able to just run ```python slowmovie.py``` and have it load the test file with the default driver. This setup would allow more advanced users to either give an argument like `-e waveshare.epd2in9` or `-e inky.impression` for displays we support through dynamic loading. People with coding skills could then easily add additional types by just adding a class for their device; hopefully over time we'd get more and more into the repo through PRs. 

I think this would keep with the original thought of sticking to Tom's core idea but allow a framework for expansion.

robweber:
Been hacking away at this a little more. I like where it is going, at least in terms of abstraction for multiple devices. I ordered an Inky device for testing to try and implement another type. I wanted a second for a test setup anyway. 

One thing I'm wondering is if this gets split off into it's own project or is it worth keeping it all together? Something like `SlowMovieEpd`. I can think of some pros and cons to that approach. 

### Pros Of Different Repo

Keeps with this repo's original vision - provide an implementation of a VSMP
Simplifies codebase here to that goal only
Provides option to keep driver loading the way it is or use this abstraction method (in a branch)

### Cons

Would need to modify instructions to build another library
Two places to maintain if issues
May "hardcode" this project to the existing display driver - lots of forks for every additional variant

I'm fine either way. I think this is a great idea but if not the right fit for the direction of this project that's ok too. Not every feature is going to be an included feature. Thoughts?

PR #38, opened 2021-04-05
Code Maintenance Housekeeping

robweber:
As we get close to merging in some of the code changes I thought we could get this into `master` first if that's ok. Basically some housekeeping stuff. 

__Issue Templates__
Added some basic templates for bugs, features, and informational requests. Should help with some of the Issues going forward. The informational ones I added mainly due to all the Issues that are just links to other forks or similar projects. We could go ahead and start posting those to like "Links" page on the wiki or something? 

__Contributing__
I went through a bunch of contributing files and pieced together something I thought made sense. With multiple author changes being merged in I think it's fair there are no surprises to people on the process for things. It's also just something we can stand behind going forward. 

__License__
I know this is in one of the other PRs but again, I think it's important this exist prior to merging in other stuff. No surprises regarding how this code is being maintained when contributing. 

This has no code modifications so I think we could merge this now. If there are any wording changes let me know.

PR #42, opened 2021-04-06
Add contributing in readme

qubist:
As mentioned in [a comment](https://github.com/TomWhitwell/SlowMovie/pull/38#pullrequestreview-627903004) on #38. 

This should be a pretty uncontroversial change. Just kicking off the practice of using a new branch and PR to make changes so we can discuss and approve each other's work.


robweber (approved):
Looks good to me. I agree it's a pain for small changes but on the flip side it negates any hurt feelings later when someone sees a change and goes "how did that get in there".

qubist:
Yup! Pretty easy to say "ok cool" on someone's change. That's a good price to pay for the opportunity to say "hold on a sec, what about ___" and then start a conversation.

PR #43, opened 2021-04-06
Add .mkv support

qubist:
* Changes video_info function to return an dictionary.
* Get framerate and frame count in a more robust way from the ffmpeg 
probe. This method works for the mkv I was testing on.
* The dictionary returned from video_info now includes duration, setting 
up for future runtime estimates.

Hopefully equivalent to f450a90, baa0526, and 8f0c58a

### DIFF
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -72,6 +72,7 @@ def video_info(file):
     stream = probeInfo['streams'][0]
 
     # Calculate framerate
+    r_fps = stream['r_frame_rate']
     fps = float(Fraction(r_fps))
 
     # Calculate duration
@@ -123,7 +124,7 @@ if not os.path.isdir(viddir):
 # First we try the file argument...
 currentVideo = args.file
 
-# ...then a random video, if selected... 
+# ...then a random video, if selected...
 if not currentVideo and args.random_file:
     videos = list(filter(supported_filetype, os.listdir(viddir)))
     if videos:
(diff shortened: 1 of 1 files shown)


slowmovie.py
```diff
@@ -50,11 +50,11 @@ def generate_frame(in_filename, out_filename, time):
         .run(capture_stdout=True, capture_stderr=True)
     )
 
-def check_mp4(value):
+def check_vid(value):
     if not os.path.isfile(value):
         raise configargparse.ArgumentTypeError("File '%s' does not exist" % value)
     if not supported_filetype(value):
-        raise configargparse.ArgumentTypeError("'%s' is not a supported file type" % value)
+        raise argparse.ArgumentTypeError(f"File '{file}' should be a file with one of the following supported extensions: {', '.join(fileTypes)}")
```
robweber, line 57: `configargparse.ArgumentTypeError` and `argparse.ArgumentTypeError` are basically the same error by design. We're using one in line 55 and a different one in 57. Should we pick one to be consistent?

robweber (commented):
Other than that one comment looking good.

slowmovie.py
qubist: Oops, yes! Copy-paste error. Let's stick with configparse as that's what's imported. Fixed in bb4db81

missionfloyd (approved):
Looks good

PR #49, opened 2021-04-07
Remember video info for each video

missionfloyd:
This is something I previously implemented in `helloworld.py`. 

The first time a given video is scanned, the results are added to a dict. Then, if the same file is encountered again, the video info is retrieved from the dict rather than running ffprobe again, which can take a few seconds on slower devices.

This'll be particularly useful if a "random frames from folder" mode is added (this is what `helloworld.py` does).

### DIFF
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -105,13 +105,14 @@ def video_info(file):
         frameTime = 1000 / fps
 
         info = {
-            'frame_count' : frameCount,
-            'fps' : fps,
-            'duration' : duration,
-            'frame_time' : frameTime }    
+            'frame_count': frameCount,
+            'fps': fps,
+            'duration': duration,
+            'frame_time': frameTime}
 
         videoInfos[file] = info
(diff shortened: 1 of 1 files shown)


slowmovie.py
```diff
@@ -181,6 +187,7 @@ def video_info(file):
 width = epd.width
 height = epd.height
 
+videoInfos = {}
```
qubist, line 190: What's this line for?

missionfloyd: That's a global variable to keep the video info for each scanned file. It just has to be somewhere before the first `video_info()` call.

robweber (approved):
I like it. No reason to keep probing a video on each update when we know the valid information.

slowmovie.py
qubist: Ah, I see now! Thanks for the explanation 👍

qubist (approved):

PR #50, opened 2021-04-07
Update Readme with service management instructions

### LINKED ISSUE #47: service fails to start
I can't get the service to auto-start on my setup. I followed these instructions in the README

If you want SlowMovie to start automatically when the device is powered on:
* `sudo cp slowmovie.service /etc/systemd/system`
* `sudo systemctl daemon-reload`
* `sudo systemctl enable slowmovie`

But I see "failed" next to slowmovie when running `sudo systemctl` to check out what's going on. Am I just doing something wrong? I haven't used services before.

qubist:
PR in response to conversation in #47.

This is a first go at it, may need some corrections, esp by @missionfloyd who knows this stuff well.

Since this is a part of the system that unexperienced users will need to interact with a bit, I tried to make instructions pretty simple and low-level, at the expense of being a bit verbose. My thought was that it's appropriate in this case to make things easy for non-hackers to set up a VSMP.

resolves #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,8 +92,8 @@ config file values, which in turn override defaults.
 SlowMovie can run as a service. To set this up, from the SlowMovie directory run the following:
 
 ```
-$ sudo cp slowmovie.service /etc/systemd/system
-$ sudo systemctl daemon-reload
+sudo cp slowmovie.service /etc/systemd/system
+sudo systemctl daemon-reload
 ```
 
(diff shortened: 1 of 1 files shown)


missionfloyd (approved):
Looks alright to me. Maybe remove the `$`'s. It's not used anywhere else in this. I could see some people thinking that's part of the command, then it doesn't work.

robweber (approved):

PR #51, opened 2021-04-07
Install Script

robweber:
Alot of the questions lately have to do with steps for the install; plus it's just nice to have a "one liner" you can use to update things. This is my first pass at an `install.sh` script. I'm not the best at bash scripting so if this can be made better shout it out. 

## Basic Parts

There are 3 basic parts: 

1. Install/Update
2. Install Service
3. Uninstall Service

### Install/Update

This follows the steps from the README document. If any of the repos already exist they'll just be updated. There is a special `-s` flag you can use to skip the dependency installs since they take a lot of time. 

### Install/Uninstall Service

Again, this just follows the readme. One installs the service files and enables them, the other removes. 

## Testing

To test this you'll have to clone this branch and run from there with some special arguments. The `install.sh` script has been updated to point to the master of this repo - which obviously doesn't contain this file yet. As such following the instructions will fail until this is merged. As a workaround for testing this is what I would advise. 

1. If you have a cloned repo, rename it so you don't lose anything. `mv SlowMovie SlowMovie2`
2. Run this command to download the script from my branch: `curl -O https://raw.githubusercontent.com/robweber/SlowMovie/install_script/Install/install.sh`
3. Give it execute permissions `chmod +x install.sh`
4. Run with args to use my branch instead of the main one `./install.sh -r https://github.com/robweber/SlowMovie -b install_script`

All of this will be unnecessary once it's merged but you have to override the defaults for testing.


Install/install.sh
```diff
@@ -0,0 +1,205 @@
+#!/bin/bash -e
+
+GIT_REPO=https://github.com/TomWhitwell/SlowMovie
+GIT_BRANCH=master
+PYTHON_VERSION=3
+SKIP_DEPS=false
+
+function install_linux_packages(){
+  sudo apt-get update
```
missionfloyd, line 9: Should the sudos be in here, or should the script itself be run with sudo?

Install/install.sh
```diff
@@ -0,0 +1,205 @@
...
+}
+
+function install_python_packages(){
+  # set the python3 as the default version
+  sudo update-alternatives --install /usr/bin/python python /usr/bin/python3 1
+  sudo update-alternatives --set python /usr/bin/python${PYTHON_VERSION}
+
+
+  sudo pip3 install setuptools -U
+  sudo pip3 install -r $LOCAL_DIR/Install/requirements.txt
```
missionfloyd, line 21: I think the python dependencies can be installed locally with [venv](https://docs.python.org/3/tutorial/venv.html), but I'm not familiar with doing it that way.

But the way we've been doing it is probably simpler.

Install/install.sh
```diff
@@ -0,0 +1,205 @@
...
+
+# clear screen
+  clear;
+
+  # Set color of logo
+  # Logo generated from: http://patorjk.com/software/taag/
+  tput setaf 4
+  tput bold
+
+  cat << EOF
```
missionfloyd, line 166:
```suggestion
  cat << "EOF"
```
Doesn't display right without the quotes.
```
pi@pizero:~ $ bash logo.sh
   _____ _               __  __            _
  / ____| |             |  \/  |          (_)
 | (___ | | _____      _| \  / | _____   ___  ___
  \___ \| |/ _ \ \ /\ / / |\/| |/ _ \ \ / / |/ _   ____) | | (_) \ V  V /| |  | | (_) \ V /| |  __/
 |_____/|_|\___/ \_/\_/ |_|  |_|\___/ \_/ |_|\___|

pi@pizero:~ $ bash logo.sh
   _____ _               __  __            _
  / ____| |             |  \/  |          (_)
 | (___ | | _____      _| \  / | _____   ___  ___
  \___ \| |/ _ \ \ /\ / / |\/| |/ _ \ \ / / |/ _ \
  ____) | | (_) \ V  V /| |  | | (_) \ V /| |  __/
 |_____/|_|\___/ \_/\_/ |_|  |_|\___/ \_/ |_|\___|
```

qubist:
Next week I'll be building a new VSMP on RaspPi 4 B from NOOBS—I'll test this on that.

README.md
```diff
@@ -141,4 +160,4 @@ Please read our [contributing guidelines](/.github/CONTRIBUTING.md) before submi
 
 ## License
 
-[MIT](/LICENSE)
```
qubist, line 144: What's happening here? Unintended whitespace change? intended whitespace correction?

Install/install.sh
robweber: I had an internal debate about that as I was doing it too. At least giving the user a heads up that we're using elevated commands is probably best practice. However, on the Rpi - the only platform we're targeting; the `pi` user gets sudo access without even a password by default anyway. For a hobbyist project we're trying to make easy how important is that - idk? I've written scripts that do other stuff where if you're not running as sudo it kicks out the script and asks you to re-run with privileges. That could be an option. 

I tried to find some examples - and this is by no means an exhaustive list - but two just put sudo right in the script and the other specifically says "run as sudo" in the instructions. These projects are all designed to work on an Rpi so figured they were good examples. What are you most comfortable with? I think we an justify either method. 

https://github.com/home-assistant/supervised-installer/blob/master/installer.sh
https://get.pimoroni.com/inky
https://github.com/home-assistant/supervised-installer/blob/master/installer.sh

README.md
robweber: That was my guess. Didn't seem to hurt anything so I left it. Could be a windows vs linux line ending thing.

Install/install.sh
robweber: Nice. I was having mixed results with it and wondered if it was my version of Putty or something. That fixed it up.

Install/install.sh
qubist: If all other things are equal, I think leaving sudo out and requiring the script to be run as sudo is cleaner.

README.md
qubist: Cool, I'm sure it can't hurt, just weird :)

README.md
```diff
@@ -20,6 +20,25 @@ SlowMovie is the code that runs a VSMP on a Raspberry Pi.
...
+
+```
+
+bash <(curl https://raw.githubusercontent.com/TomWhitwell/SlowMovie/master/Install/install.sh)
+
+```
+
+You can re-run the script at any time in the `Install` folder to update.
+
+### Manually
```
qubist, line 40: Just some tweaks to make this a bit clearer hopefully.

Note that I couldn't use the triple-backtick format for code blocks while doing one of these review suggestions, so I had to change it to the quadruple-spaces formatting. It makes the same thing when rendered but it would be nice to keep it looking like
\```
bash .....
\```
In the actual change for consistency with the rest of the readme.

```suggestion
### Automated installation

You can quickly install this repository and all required libraries via an install script. This is a simple way to get started if you're not as comfortable with the command line.

To run the install script, open your terminal, copy-paste the following command in, and hit enter.

    bash <(curl https://raw.githubusercontent.com/TomWhitwell/SlowMovie/master/Install/install.sh)

You'll be presented with 4 options when you run the script:

1. **Install/Upgrade SlowMovie** - download the repository and install/update any libraries needed
2. **Install SlowMovie Service** - run the commands install the SlowMovie service file as described below
3. **Uninstall SlowMovie Service** - uninstall the SlowMovie service
4. **Exit**

You can re-run the script at any time in the `Install` folder to update.

### Manual installation
```

Install/install.sh
robweber: Playing around with this a bit more removing `sudo` from the script itself will make the install instructions a little more complicated looking. We'd have to do something like `sudo su -c "bash <(curl http://path/to/install.sh)" root`. It looks a lot _scarier_. An alternative would be to have the user download it first and then run `sudo ./install.sh`. 

In testing another command that would work is `curl http://path/to/install.sh | sudo bash` but that would only work for non-interactive scripts. It looks a lot more user friendly though. I could modify this for to run without prompts but we'd have to make some compromises on steps 2 and 3 (flags perhaps?)

robweber:
@qubist - if you're going to run this as part of a new install you can use [this branch](https://github.com/robweber/SlowMovie/tree/install_script_testing). I modified the install script to point at this repo so all the pieces will work without a lot of hoopla. It's updated to the current master - I'll try to mirror it for the next week or so.  You can use this command to start it off: 

```
bash <(curl https://raw.githubusercontent.com/robweber/SlowMovie/install_script_testing/Install/install.sh)
```

When you're all done go to the SlowMovie directory and run `git remote set-url origin https://github.com/TomWhitwell/SlowMovie.git` to swap the repo back to the original for future updates. I've tested this by deleting all my local folders but not on a clean OS install yet (assuming our instructions work!).

Install/install.sh
qubist: Interactivity and a friendly-looking command is probably more important than removing sudos from the install script. If it keeps the command simple and interactivity, I say it's worth it to have sudos in the script, especially if you found other examples where folks do that

missionfloyd: That makes sense.

robweber: The only other way I could see it working, and stay with a one line install idea, is to use more flags. Something like 

```
# just do the base install, but install service by default as well
curl http://path/to/install.sh | sudo bash
```

```
# uninstall service
sudo ./install -u
```
That would still work but I'd include the service install by default and remove the menu. I don't think adding more instructions to re-run just for that will be the best user engagement either. A separate flag to uninstall if someone wanted that would be available. I think either way is doable but we will need to make some "onboarding vs usability" trade offs.

Install/install.sh
missionfloyd, line 21: requirements.txt isn't strictly necessary.
```suggestion
  sudo pip3 install ffmpeg-python pillow ConfigArgParse
```
This works, too. The install directory wouldn't be necessary either, as install.sh would be the only thing in it.

Either way's fine, though.

Install/install.sh
```diff
@@ -0,0 +1,205 @@
...
+    echo -e "Updating Waveshare Drivers"
+    cd $WAVESHARE_DIR
+    git fetch
+    git pull
+  else
+    echo -e "Installing Waveshare Drivers"
+    git clone https://github.com/waveshare/e-Paper
+  fi
+
+  echo -e "$(tput setaf 1)Be patient - this takes time$(tput sgr 0)"
```
missionfloyd, line 37: No need to use tput. Colored text can be accomplished with [control codes](https://techstop.github.io/bash-script-colors/).
```
echo -e "\e[0;31mBe patient - this takes time\e[0m"
```

Install/install.sh
robweber: I thought about that but it affects the ability of using the script for both and install and upgrade. If they python libraries were to change in some way the script wouldn't work for an upgrade. The user would run the script from the `SlowMovie` directory and it would update the repo but because it's running the script in memory the python libs wouldn't change. You'd have to re-run the script after the repo update to get new python libraries installed. With the requirements file you can update the repo files and then update the python libraries in the same session. 

Perhaps renaming this to `setup.sh` is more intuitive that it's for both purposes?

Install/install.sh
robweber: I've used those before but in future looks always was irritated it wasn't easier to know what code = what color. I like the example in that link though, just set some variables for the colors.

Install/install.sh
qubist: Oh that's really clever:
* The script first downloads the repo, which includes possibly new code and the requirements for that new code.
* Then it installs the requirements that may have just been downloaded.

Real sweet. IMO, `install.sh` is clearer than `setup.sh` even though it can do both install and update

Install/install.sh
```diff
@@ -0,0 +1,205 @@
...
+  if [ -f "/etc/systemd/system/slowmovie.service" ]; then
+    # stop if running and remove service files
+    sudo systemctl stop slowmovie
+    sudo systemctl disable slowmovie
+    sudo rm /etc/systemd/system/slowmovie.service
+    sudo systemctl daemon-reload
+
+    echo -e "SlowMovie service is uninstalled"
+  else
+    echo -e "$(tput setaf 1)SlowMovie is not installed, nothing to clean up$(tput sgr 0)"
```
qubist, line 127:
```suggestion
    echo -e "SlowMovie service was successfully uninstalled"
  else
    echo -e "$(tput setaf 1)SlowMovie service is already uninstalled.$(tput sgr 0)"
```

Install/install.sh
```diff
@@ -0,0 +1,205 @@
...
+echo "2 - Install SlowMovie Service"
+echo "3 - Uninstall SlowMovie Service"
+echo "4 - Exit"
+echo ""
+
+INSTALL_OPTION=0
+
+while [ "$INSTALL_OPTION" = "" ] || [ $INSTALL_OPTION -le 0 ] || [ $INSTALL_OPTION -gt 4 ]; do
+        read -p "Choose what you want to do? " INSTALL_OPTION
+done
```
missionfloyd, line 194:
```suggestion
while [[ "$INSTALL_OPTION" != "" ]] && [[ $INSTALL_OPTION != [1-4] ]]; do
        read -p "Choose what you want to do? " INSTALL_OPTION
done

: ${INSTALL_OPTION:=1}
```
This will default to 1 if left blank. Using `-le` and `-gt` was also causing it to crash if anything but numbers were entered.

Install/install.sh
```diff
@@ -0,0 +1,205 @@
...
+echo "1 - Install/Upgrade SlowMovie"
+echo "2 - Install SlowMovie Service"
+echo "3 - Uninstall SlowMovie Service"
+echo "4 - Exit"
+echo ""
+
+INSTALL_OPTION=0
+
+while [ "$INSTALL_OPTION" = "" ] || [ $INSTALL_OPTION -le 0 ] || [ $INSTALL_OPTION -gt 4 ]; do
+        read -p "Choose what you want to do? " INSTALL_OPTION
```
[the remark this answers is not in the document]
missionfloyd: Many programs put the choice in brackets.

    Do you want to continue? [Y/n]

Or, could make menus with dialog.
```
dialog --menu "Choose what you want to do." 0 0 0 1 "Install/Upgrade SlowMovie" 2 "Install SlowMovie Service" 3 "Uninstall SlowMovie Service"
```
Untitled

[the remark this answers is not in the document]
qubist: Does `INSTALL_OPTION=` install that automatically? Seems like a robust option as we're basically trying to create a simple menu.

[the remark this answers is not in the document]
missionfloyd: It outputs to a variable just like `read` does.

Install/install.sh
```diff
@@ -0,0 +1,205 @@
...
+  WAVESHARE_DIR=/home/pi/e-Paper
+
+  if [ -d "${WAVESHARE_DIR}" ]; then
+    echo -e "Updating Waveshare Drivers"
+    cd $WAVESHARE_DIR
+    git fetch
+    git pull
+  else
+    echo -e "Installing Waveshare Drivers"
+    git clone https://github.com/waveshare/e-Paper
```
missionfloyd, line 34:
```suggestion
    git clone --depth 1 https://github.com/waveshare/e-Paper
```
Only current revision. Faster, slightly less disk space.

Install/install.sh
```diff
@@ -0,0 +1,205 @@
...
+  else
+    echo -e "No Install Found - Cloning Repo"
+    git clone ${GIT_REPO} ${LOCAL_DIR}
+  fi
+
+  # update the repo
+  cd $LOCAL_DIR
+  git fetch
+  git checkout $GIT_BRANCH
+  git pull
```
missionfloyd, line 81:
```suggestion
  # update the repo
  cd $LOCAL_DIR
  git pull
else
  echo -e "No Install Found - Cloning Repo"
  git clone ${GIT_REPO} ${LOCAL_DIR}
  fi
```
Only `git pull` is needed to update, and there's no point pulling immediately after cloning.

README.md
```diff
@@ -20,6 +20,25 @@ SlowMovie is the code that runs a VSMP on a Raspberry Pi.
...
+    bash <(curl https://raw.githubusercontent.com/TomWhitwell/SlowMovie/master/Install/install.sh)
+
+You'll be presented with 4 options when you run the script:
+
+1. **Install/Upgrade SlowMovie** - download the repository and install/update any libraries needed
+2. **Install SlowMovie Service** - run the commands install the SlowMovie service file as described below
+3. **Uninstall SlowMovie Service** - uninstall the SlowMovie service
+4. **Exit**
+
+You can re-run the script at any time in the `Install` folder to update.
```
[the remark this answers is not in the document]
missionfloyd: whiptail --yesno "SlowMovie install complete. To test, run './slowmovie.py'\n\nWould you like to install SlowMovie service to\nstart playback automatically?" 0 0
Then check the value of `$?`. 0 is Yes, 1 is No

Install/install.sh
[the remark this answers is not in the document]
qubist: Yeah, I misunderstood at first but now got it. Seems to me like using whiptail would be better since it's already installed. Overall I like this option

Install/install.sh
```diff
@@ -0,0 +1,205 @@
...
+function install_linux_packages(){
+  sudo apt-get update
+
+  sudo apt-get install -y ffmpeg git python3-pip
+}
+
+function install_python_packages(){
+  # set the python3 as the default version
+  sudo update-alternatives --install /usr/bin/python python /usr/bin/python3 1
+  sudo update-alternatives --set python /usr/bin/python${PYTHON_VERSION}
```
missionfloyd, line 17: Are you sure we should be changing this?

Install/install.sh
robweber: I want to keep the `fetch` and the `checkout` as it is possible to switch branches. You could envision a scenario where someone might be running off a branch and then want to switch to a different one. I did move some things around to checkout the branch specifically with `git clone -b` and then move the other commands to within the IF statement.

Install/install.sh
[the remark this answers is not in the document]
robweber: I'll be honest I'm not super crazy about that type of menu but it's a small concession of that's the group consensus. Implemented now.

Install/install.sh
robweber: Not really needed anymore with the `whiptail` menus

Install/install.sh
robweber: Yes.

missionfloyd: Doesn't seem necessary since we're explicitly running `python3`. Some programs still expect `/usr/bin/python` to point to python 2, and in Debian Bullseye it's being removed altogether in favor of `python2` and `python3`.

And I don't think we should be altering other installed programs,
```suggestion
```

Install/install.sh
```diff
@@ -0,0 +1,192 @@
...
+
+# set the local directory
+LOCAL_DIR="/home/pi/$(basename $GIT_REPO)"
+
+echo -e "SlowMovie Repo set to ${YELLOW}${GIT_REPO}/${GIT_BRANCH}${RESET}"
+echo -e "Setting up in local directory ${YELLOW}${LOCAL_DIR}${RESET}"
+echo -e ""
+cd /home/pi/
+
+INSTALL_OPTION=$(whiptail --menu "Choose what you want to do." 0 0 0 1 "Install/Upgrade SlowMovie" 2 "Install SlowMovie Service" 3 "Uninstall SlowMovie Service" 4 "Exit" 3>&1 1>&2 2>&3)
```
missionfloyd, line 169:
```suggestion
cd /home/pi/

INSTALL_OPTION=$(whiptail --menu "\
   _____ _               __  __            _      
  / ____| |             |  \/  |          (_)     
 | (___ | | _____      _| \  / | _____   ___  ___ 
  \___ \| |/ _ \ \ /\ / / |\/| |/ _ \ \ / / |/ _ \\
  ____) | | (_) \ V  V /| |  | | (_) \ V /| |  __/
 |_____/|_|\___/ \_/\_/ |_|  |_|\___/ \_/ |_|\___|

 Repo set to '${GIT_REPO}/${GIT_BRANCH}'
 Setting up in local directory '${LOCAL_DIR}'

 Choose what you want to do." 0 0 0 \
1 "Install/Upgrade SlowMovie" \
2 "Install SlowMovie Service" \
3 "Uninstall SlowMovie Service" \
3>&1 1>&2 2>&3)
```

This still fits.

Install/install.sh
```diff
@@ -0,0 +1,192 @@
...
+
+: ${INSTALL_OPTION:=4}
+
```
missionfloyd, line 172:
```suggestion
```
Don't need this anymore.

Install/install.sh
```diff
@@ -0,0 +1,192 @@
...
+if [ $INSTALL_OPTION -eq 1 ]; then
+	# install or update
+  install_slowmovie
+
+  # prompt for service install if the first time being run
+  if [ $? -eq 0 ]; then
+    whiptail --yesno "SlowMovie install complete. To test, run 'python3 slowmovie.py'\n\nWould you like to install the SlowMovie Service to\nstart playback automatically?" 0 0
+
+    if [ $? -eq 0 ]; then
```
missionfloyd, line 181:
```suggestion
    if (whiptail --yesno "SlowMovie install complete. To test, run 'python3 slowmovie.py'\n\nWould you like to install the SlowMovie Service to\nstart playback automatically?" 0 0); then
```
Apparently this can be simplified.

Install/install.sh
robweber: I guess agree to disagree? [Python2 is EOL](https://www.python.org/doc/sunset-python-2/) and helping it limp along isn't doing anyone any favors.  The target for this isn't some high-end enterprise server with 10 other apps on it. It's a small device that is probably going to do this, and this only. This change is mostly meant to save users from themselves. I'll pull it but get ready for some questions about "I ran your script and `python slowmovie.py` throws an error when people attempt to run it that way.

missionfloyd: I agree that everyone should move on. Python 2 is being [depricated](https://wiki.debian.org/Python/2Removal) soon, and running `python` won't do anything (When that happens, the symlink can be resored for compatibility with the [`python-is-python2`](https://packages.debian.org/bullseye/python-is-python2) and [`python-is-python3`](https://packages.debian.org/bullseye/python-is-python3) packages)

The instructions do way to run `python3 slowmovie.py`, but maybe `./slowmovie.py' would be better to avoid confusion. It is marked executable.

Install/install.sh
```diff
@@ -0,0 +1,189 @@
...
+    echo -e "Skipping dependency installs, updating SlowMovie code only"
+  fi
+
+  if [ -d "${LOCAL_DIR}" ]; then
+    echo -e "Existing Install Found - Updating Repo"
+    cd $LOCAL_DIR
+    git fetch
+    git checkout $GIT_BRANCH
+    git pull
+
```
missionfloyd, line 80:
```suggestion

# Update service if installed and file changed
if [ -f "/etc/systemd/system/slowmovie.service" ] && ! (cmp -s "slowmovie.service" "/etc/systemd/system/slowmovie.service"); then
  sudo cp slowmovie.service /etc/systemd/system
  sudo systemctl daemon-reload
fi

```
I don't see slowmovie.service changing any more, but just in case.

Install/install.sh
robweber: Going to keep it, if they hit __cancel__ the variable comes back blank.

Install/install.sh
robweber: I ended up with a variation of this. The prompt is now prior to the function call so that we have the user's attention. I don't bother with any of the other functions or checks. There is only so much polish we need for an install script.

Install/install.sh
robweber: I did add this but made [some modifications](https://github.com/robweber/SlowMovie/blob/install_script/Install/install.sh#L112).

README.md
```diff
@@ -20,6 +20,25 @@ SlowMovie is the code that runs a VSMP on a Raspberry Pi.
...
+
```
qubist, line 39:
```suggestion
You'll be presented with 4 options when you run the script:

1. **Install/Upgrade SlowMovie** - download the repository and install/update any libraries needed
2. **Install SlowMovie Service** - run the commands install the SlowMovie service file as described below
3. **Uninstall SlowMovie Service** - uninstall the SlowMovie service
4. **Exit**

For first-time automated installation, choose 1: Install/Upgrade SlowMovie. When prompted, you can choose "yes" to have the SlowMovie service installed as well which will enable playback to start automatically when the device is powered on or rebooted.

```

Little change to make this a bit more walkthrough-y and removing update instructions to somewhere else.

qubist: I think the

```
You can re-run the script at any time in the `Install` folder to update.
```

that I took out should be moved to a new subheading Usage -> Updating but it's not letting me suggest down there.

qubist (changes requested):
Also I think the `Running as a service` section needs a bit of an update because of the new install script. Heading should change to `SlowMovie service` or something, and the content should basically be:
* To install it either:
    * use the install script, or
    * do the manual process (what's already there)
* to use it:
    * (stuff that's already there)

I think it's not letting me suggest in that area because there weren't changes there

Install/install.sh
missionfloyd: Yeah, I thought that might be overkill.

Install/install.sh
missionfloyd: I forgot that -eq only works on numbers.

missionfloyd:
Perhaps a display menu would be appropriate.
(it'd be nice if waveshare displays had auto-detection like pimoroni's)
```
DISPLAY=$(whiptail --notags --title "Select your display" --menu "" 0 0 0 \
epd1in02     "1.02-inch (120x80)" \
epd1in54     "1.54-inch (200x200)" \
epd1in54_V2  "1.54-inch V2 (200x200)" \
epd1in54b    "1.54-inch Red (200x200)" \
epd1in54b_V2 "1.54-inch Red V2 (200x200)" \
epd1in54c    "1.54-inch Yellow (152x152)" \
epd2in13     "2.13-inch (250x122)" \
epd2in13_V2  "2.13-inch V2 (250x122)" \
epd2in13bc   "2.13-inch Red/Yellow (250x122)" \
epd2in13b_V3 "2.13-inch Red/Yellow V2 (250x122)" \
epd2in13d    "2.13-inch Flexible (212x104)" \
epd2in66     "2.33-inch (296x152)" \
epd2in66b    "2.33-inch Red (296x152)" \
epd2in7      "2.7-inch (264x176)" \
epd2in7b     "2.7-inch Red (264x176)" \
epd2in7b_V2  "2.7-inch Red V2 (264x176)" \
epd2in9      "2.9-inch (296x128)" \
epd2in9_V2   "2.9-inch V2 (296x128)" \
epd2in9bc    "2.9-inch Red/Yellow (296x128)" \
epd2in9b_V3  "2.9-inch Red/Yellow V3 (296x128)" \
epd2in9d     "2.9-inch Flexible (296x128)" \
epd3in7      "3.7-inch (480x280)" \
epd4in01f    "4.01-inch Color (640x400)" \
epd4in2      "4.2-inch (400x300)" \
epd4in2bc    "4.2-inch Red/Yellow (400x300)" \
epd4in2b_V2  "4.2-inch Red/Yellow V2 (400x300)" \
epd5in65f    "5.65-inch Color (880x528)" \
epd5in83     "5.83-inch (600x448)" \
epd5in83_V2  "5.83-inch V2 (648x480)" \
epd5in83bc   "5.83-inch Red/Yellow (600x448)" \
epd5in83b_V2 "5.83-inch Red/Yellow V2 (648x480)" \
epd7in5      "7.5-inch (640x384)" \
epd7in5_V2   "7.5-inch V2 (800x480)" \
epd7in5bc    "7.5-inch Red/Yellow (640x384)" \
epd7in5b_V2  "7.5-inch Red V2 (800x480)" \
epd7in5_HD   "7.5-inch HD (880x528)" \
epd7in5b_HD  "7.5-inch HD Red (880x528)" \
3>&2 2>&1 1>&3)
```
Then driver could be set in slowmovie.py (or slowmovie.conf if implemented) with sed.

missionfloyd:
> I think it's not letting me suggest in that area because there weren't changes there

There's an issue open for this https://github.com/isaacs/github/issues/1655

robweber:
I like the idea of supporting more than one display but I don't think doing it as part of the install file is the right method for that. It would still limit this to only waveshare displays and would require the install script modifying the python file prior to use to insert the right import. What about people that don't use the install script. The python files should run on their own, with sane defaults. 

Dynamically importing the right library at runtime is probably the best method of attaining this. The user could still specify a device type but `slowmovie.py` would import based on that.

README.md
robweber: I'll also add the other changes you mentioned in separate commit.

README.md
```diff
@@ -90,9 +109,9 @@ If an arg is specified in more than one place, then commandline values override
 config file values, which in turn override defaults.
 ```
 
-### Running as a service
+### SlowMovie Service
 
-SlowMovie can run as a service. To set this up, from the SlowMovie directory run the following:
+SlowMovie can run as a service. To set this up you can either use option 2 from the install script ([see above](https://github.com/TomWhitwell/SlowMovie/#automated-installation)) or from the SlowMovie directory run the following:
```
qubist, line 114: Awesome

qubist (commented):
Looks great! This seems like it's becoming finalized. I'm going to run it during setup on my second rpi today and add any notes.

qubist:
Just tested this on a fresh rpi:

[asciicast](https://asciinema.org/a/1brUHYXdd8YR8tAFCqSyuZve9)

(click for video)

Worked great but I think it was an outdated version? I didn't get the prompt to install the service, but it seemed like it installed in anyway. Was that expected behavior?

Install/install.sh
```diff
@@ -0,0 +1,243 @@
...
+  # check if the service file needs to be updated
+  if (service_installed) && ! (cmp -s "slowmovie.service" "/etc/systemd/system/slowmovie.service"); then
+    sudo cp $SERVICE_FILE $SERVICE_DIR
+    sudo systemctl daemon-reload
+
+    echo -e "Updating SlowMovie service file"
+  fi
+
+  echo -e "SlowMovie install/update complete. To test, run '${YELLOW}python3 slowmovie.py${RESET}'"
```
qubist, line 119: This should have the path to the slowmovie.py file in it, say "from the SlowMovie directory, run..."

When running the installer I ran it from `~` and if I didn't know better, running `python3 slowmovie.py` from there wouldn't work.

qubist (changes requested):

robweber:
I think you found a but in the install script regarding the prompt for the service install. I noticed on the screen capture there was an error `line 220: [: missing ]`. That's where the if statement starts to do the check for a previous install. Something wrong with the brackets there. I think it installed anyway as the option is set to `1` which evals to True later on. The default should be `0` so that on upgrades it evals to False. 

Glad we had a real world test of this thing.

qubist:
Yeah, it was a good catch! Do you have enough info from what I posted to fix the issue?

robweber:
Yea I think so. Thanks for doing the details video.

missionfloyd:
> I think you found a but in the install script regarding the prompt for the service install. I noticed on the screen capture there was an error `line 220: [: missing ]`. That's where the if statement starts to do the check for a previous install. Something wrong with the brackets there. I think it installed anyway as the option is set to `1` which evals to True later on. The default should be `0` so that on upgrades it evals to False.

Double the brackets `[[ ! -d "${LOCAL_DIR}" ]]`

Install/install.sh
robweber: Added with c84cb168f9094cbce8a3371ac2222f2b8c29afbf

Install/install.sh
```diff
@@ -116,7 +116,7 @@ function install_slowmovie(){
     echo -e "Updating SlowMovie service file"
   fi
 
-  echo -e "SlowMovie install/update complete. To test, run '${YELLOW}python3 slowmovie.py${RESET}'"
+  echo -e "SlowMovie install/update complete. To test, run '${YELLOW}python3 ${LOCAL_DIR}/slowmovie.py${RESET}'"
```
missionfloyd, line 119: Slowmovie.py is executable, `python3` isn't needed.

missionfloyd:
pip can apparently install directly from github
```
pip3 install 'git+https://github.com/waveshare/e-Paper.git#egg=waveshare-epd&subdirectory=RaspberryPi_JetsonNano/python'
```
It can also go in requirements.txt

robweber:
Added the waveshare libs to the `requirements.txt` file. That will simplify the build a lot. I didn't go through the effort yet but probably worth updating the README for the manual instructions too right? Step 2 could be reduced to: 


2. Install Waveshare e-paper drivers
  * `sudo pip3 install git+https://github.com/waveshare/e-Paper.git#egg=waveshare-epd&subdirectory=RaspberryPi_JetsonNano/python`

qubist:
Does the pip3 install of the GitHub repo run `python3 setup.py install` for us?

robweber:
Where are we with this? I think everything is taken care of unless I'm missing something?

qubist:
I can give it a final test right now just to double-check the new changes

qubist:
Okay just ran it fine from an already installed version, (which is an edge-case I guess). It was only up-to-date with 
c84cb16 because the testing repo is a bit behind. I wasn't asked about service install, I believe this is expected behavior because it's already installed?

qubist:
Manually uninstalled the service and now I get the prompt. All seeming good so far...

qubist (approved):
Everything seems tested, to work smoothly, and the documentation is updated

robweber:
One small final change. I updated the build instructions in the README to build the Waveshare drivers with the `pip` command rather than cloning and building by hand. We can revert if manual build is what we want for the manual instructions.

qubist (approved):
I like that change!

qubist:
@TomWhitwell I think since this is now merged, now would be the time for a more major change to the instructions article. Let me know how I can help!

TomWhitwell:
OK I’ll try to get to it this week

PR #53, opened 2021-04-07
Refactor code that looks at the videos directory and gets next video 

qubist:
From db61009

### DIFF
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -117,7 +117,7 @@ def video_info(file):
 # Returns the next video in the videos directory, or the first one if there's no current video
 def get_next_video(viddir, currentVideo=None):
     # Only consider videos in the directory
-    videos = list(filter(supported_filetype, os.listdir(viddir)))
+    videos = sorted(list(filter(supported_filetype, os.listdir(viddir))))
 
     # Return None if there are no videos
     if not videos:
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -84,19 +84,19 @@ def video_info(file):
         info = videoInfos[file]
     else:
         probeInfo = ffmpeg.probe(file)
-        stream = probeInfo['streams'][0]
+        stream = probeInfo["streams"][0]
 
         # Calculate framerate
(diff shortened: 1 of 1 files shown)


slowmovie.py
```diff
@@ -105,15 +105,44 @@ def video_info(file):
...
+            "frame_time": frameTime}
 
         videoInfos[file] = info
     return info
 
 
+# Returns the next video in the videos directory, or the first one if there's no current video
+def get_next_video(viddir, currentVideo=None):
+    # Only consider videos in the directory
+    videos = list(filter(supported_filetype, os.listdir(viddir)))
```
missionfloyd, line 120:
```suggestion
    videos = sorted(list(filter(supported_filetype, os.listdir(viddir))))
```
Should put this back

qubist: Thanks ✅

slowmovie.py
```diff
@@ -105,15 +105,44 @@ def video_info(file):
...
+
+    if currentVideo:
+        try:
+            nextIndex = videos.index(currentVideo) + 1
+            # If we're not wrapping around
+            if not nextIndex >= len(videos):
+                return os.path.join(viddir, videos[nextIndex])
+        except ValueError:
+            # Maybe currentVideo is no longer in the list and that's fine
+            pass
```
missionfloyd, line 134: How would it end up here? If the current video is deleted/moved/renamed while it's running, wouldn't `generate_frame()` fail?

robweber: I wonder if this was more likely in previous versions when the video directory was read at program start and then never refreshed? The chance of the video being deleted is not 0 but it's extremely small. You'd have to delete it between the generate frame call but prior to this, that window is probably measured in milliseconds.

qubist: Yeah, I think that's what happened here. I'll update the code. Thanks for catching this :)

qubist:
Ready for merge I believe, thoughts?

robweber (commented):
I like it, simplifies a lot of that code.

robweber (approved):

PR #54, opened 2021-04-08
Logging

missionfloyd:
Okay, I'll start this off. What I've got so far is logging to a file with timestamps
```
[2021-04-07 21:28:25,302] Update interval: 120
[2021-04-07 21:28:25,306] Frame increment: 4
[2021-04-07 21:28:27,395] Playing 'test.mp4'
[2021-04-07 21:28:27,398] Video info: 50 frames, 23.976fps, duration: 2.073s
[2021-04-07 21:28:27,402] This video will take 25.0 minute(s) to play.
[2021-04-07 21:28:31,361] Displaying frame 0 of test.mp4 (0.0%)

```
and printing to the console without.
```
pi@pizero:~/SlowMovie $ python3 slowmovie.py
Update interval: 120
Frame increment: 4
Playing 'test.mp4'
Video info: 50 frames, 23.976fps, duration: 2.073s
This video will take 25.0 minute(s) to play.
Displaying frame 0 of test.mp4 (0.0%)

```
However, some messages shouldn't go to the log. Something to do with log levels? (haven't used logging module before) But anything to printed to the console still goes to the journal. 

Could keep service flag and run them through an `if`, or output them to stderr instead and have the service ignore stdout (or vice-versa), or just set the service file to ignore all of it and forget about the journal.

Although [python-systemd](https://github.com/systemd/python-systemd/) has a logging handler. That's might be the right way to do it.

Thoughts? Or anything else?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,7 @@ $RECYCLE.BIN/
 
 ### SlowMovie ###
 nowPlaying
-logs/
+progress/
 Videos/*
 !Videos/test.mp4
+slowmovie.log
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -36,7 +36,7 @@ fileTypes = [".mp4", ".m4v", ".mkv"]
 
 # Handle when the program is killed and exit gracefully
 def exithandler(signum, frame):
-    logger.info('Exiting Program')
+    logger.info("Exiting Program")
     try:
         epd_driver.epdconfig.module_exit()
     finally:
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -189,7 +189,7 @@ parser.add_argument("-i", "--increment", default=defaultIncrement, type=int, hel
 parser.add_argument("-s", "--start", type=int, help="start playing at a specific frame")
 parser.add_argument("-c", "--contrast", default=defaultContrast, type=float, help="adjust image contrast (default: %(default)s)")
 parser.add_argument("-l", "--loop", action="store_true", help="loop a single video; otherwise play through the files in the videos directory")
-parser.add_argument("-o", "--loglevel", default="INFO", type=str.upper, choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"], help="minimum importance-level of messages saved to the logfile")
+parser.add_argument("-o", "--loglevel", default="INFO", type=str.upper, choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"], help="minimum importance-level of messages saved to the logfile (default: %(default)s)")
 args = parser.parse_args()
 
 # Move to the directory where this code is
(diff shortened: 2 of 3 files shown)


robweber:
Could we try and use the logging level as a flag to help determine what's shown in the log vs console? We could use `logging.info` and `logging.debug` statements in the code. When the log level is set to debug you'd see everything but at default you'd only get the info statements. Similar to the `--service`` flag idea but instead of us having to code the checks just pass that stuff to the logger and let it deal with it. 

As far as the individual log file vs the journal I guess I don't have a preference. I think one or the other is sufficient for a file based record and then we'd have console for when running interactively. I've seen lots of programs do it either way.

qubist:
The timestamps in the log vs. no timestamps in console output is a great touch.

I like using the logging handler to implement those `info` and `debug` logging levels like rob mentioned, which also keeps the code simple without duplicated if-statements.

I think I prefer using the journal, seeing as that's more universal than rolling our own logging file. At least that's what it seems like from my understanding of the situation.

missionfloyd:
Sorted out the logging levels. Using debug for things that should go to the console only.

That still leaves the journal.  By default, services send stdout to the journal. This means thousands of lines of "Displaying frame". With the log file, the journal seems redundant, so I'm leaning towards setting the service it to ignore stdout, and use the log file instead.

robweber:
What if we pass in the log level as an arg? The console and log file wouldn't need to be hard coded at info and debug. Info could be the default but if you wanted a debug log or to run the console at the info, warning, or error levels you could. 

Might be nice for bug report to ask for debug logs.

missionfloyd:
A pair of args for console level and log level? Or maybe just a `--debug` flag do dump everything to the log?

robweber:
I was typing that from my phone so didn't get too detailed. I was thinking a flag where you'd pass a level like ```--loglevel INFO``` or ```--loglevel ERROR```. This way the user could set the level depending on what they want to collect. A sane default might be INFO or WARN depending on what we want to actually see in the physical file on a normal basis. 

In the file you could translate this to the right log level setting with something like this (not sure if these are the exact attribute names). 

```
logLevels = {"INFO": logging.INFO, "ERROR": logging.ERROR, "WARN": logging.WARN, "DEBUG": logging.DEBUG} 

fileHandler.setLevel(logLevels[args.loglevel])

```

Just a `--debug` flag would be fine too if all the levels is kind of a lot. Just use an if statement to set the level in the actual handler: `logLevel = logging.INFO if not arg.debug else logging.DEBUG`

missionfloyd:
Okay, added. Feel free to fix up the help text and add a short argument if you'd like.

qubist (approved):
Love it!

missionfloyd:
Okay, I think we're ready to merge.

qubist (approved):
Let's do it

PR #55, opened 2021-04-08
Add comments and clean up code a bit

qubist:
Only 3 code changes here, all trivial. I'd appreciate a read-through of the new comments, though, to make sure they're salient!

### DIFF
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -29,16 +29,20 @@ defaultDelay = 120
 defaultContrast = 1.0
 defaultDirectory = "Videos"
 
+# Compatible video file-extensions
 fileTypes = [".mp4", ".m4v", ".mkv"]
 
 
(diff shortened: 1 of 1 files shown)


slowmovie.py
```diff
@@ -209,6 +219,7 @@ def estimate_runtime(delay, increment, frames, output="guess"):
 if not args.random_frames:
     print("Frame increment: " + str(args.increment))
```
qubist, line 220: Any thoughts on changing everything in here to fstrings? I'm happy to do it if there's no concerns, or no reason to avoid them. I think they're easier to read (and more modern I guess).

robweber (commented):
Other than the minor suggestions looks good.

qubist:
Gonna merge, we just gotta be on the lookout for little merge-conflicts later. 👍

Issue #62, opened 2021-04-14
How can I change the size of the display? I have a 2in9 display


fadihh:
**Describe the problem**
Problems should be inconsistencies with how things work or enhancements to the program. For error messages or things not working use a bug report. 

**Describe the solution you'd like**

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions that would also fix this issue.

missionfloyd:
We need to add this to the instructions.

To change the display, open `slowmovie.py` and change `epd7in5_V2` on [line 25](https://github.com/TomWhitwell/SlowMovie/blob/3c7d5c97b02ae2992e7a912252899e2ed09a2f21/slowmovie.py#L25) to the correct [driver](https://github.com/waveshare/e-Paper/tree/master/RaspberryPi_JetsonNano/python/lib/waveshare_epd). In your case, one of the following.

Driver | Display
--- | ---
`epd2in9` or `epd2in9_V2` | Black/White
`epd2in9bc` or `epd2in9b_V3` | Black/White/Red or Black/White/Yellow
`epd2in9d` | Flexible

Displays purchased recently are the newer variants (V2/V3).

PR #63, opened 2021-04-14
Issue templates

qubist:
Inspired by #62 

1. adding comments so the instructions don't show in posted issues

2. adding `help` tag and issue template
    * My thought is, until we have a Wiki or something, people are probably going to post issues asking for help with various problems. Without a help option, they'll probably just get labeled as bugs even when they're not.
    * Maybe eventually we can use a Wiki for common problems and maybe even the new Discussions feature. Something we decide works well for this type of thing. Until then, I think this'll be good to have a place for this type of activity which we've seen to be common for this repo.

### DIFF
--- a/.github/ISSUE_TEMPLATE/help.md
+++ b/.github/ISSUE_TEMPLATE/help.md
@@ -1,0 +1,16 @@
+---
+name: Help
+about: 'Need help setting up a VSMP?'
+title: ''
+labels: help
+assignees: ''
+
+---
+
+<!--- Please fill in as much as you can, the more information we have, the more we can help! --->
+
+**Describe the problem**
+<!--- Be as specific as you can, include what should happen vs what is actually happening if it's a functionality issue. --->
+
+**Platform Information** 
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/help.md
```diff
@@ -0,0 +1,16 @@
+---
+name: Help
+about: 'Need help setting up a VSMP?'
+title: ''
+labels: help
```
robweber, line 5: Could we use the __question__ label instead of __help__? Question already exists and will avoid confusion with the __help wanted__ label.

robweber (changes requested):
Looks good to me other than the one question I had. Good idea!

.github/ISSUE_TEMPLATE/help.md
qubist: Sounds great! see 43a26ef

robweber (approved):

missionfloyd (approved):

PR #66, opened 2021-04-16
Random frames from all videos

missionfloyd:
Since `helloworld.py` was meant to be an [example](https://github.com/TomWhitwell/SlowMovie/pull/29#discussion_r606103910), this simplifies it somewhat and adds it's functionality to `slowmovie.py`. Now, if both `-r` and `-R` are specified, a random frame from a random video will be chosen on every refresh, as `helloworld.py` currently does.

I've also made some other minor adjustments.
- Replaced all remaining % formatting with f-strings.
- Display video duration as HH:MM:SS
- Get framerate from `avg_frame_rate`, as `r_frame_rate` appears to be incorrect on some (variable framerate?) videos.
- Added MOV filetype. The format is similar to MP4, so it works fine as-is.
- Removed `defaultContrast` variable. [1.0 Is the default](https://pillow.readthedocs.io/en/stable/reference/ImageEnhance.html#PIL.ImageEnhance.Contrast).


slowmovie.py
```diff
@@ -95,8 +94,8 @@ def video_info(file):
         stream = probeInfo["streams"][0]
 
         # Calculate framerate
-        r_fps = stream["r_frame_rate"]
-        fps = float(Fraction(r_fps))
+        avg_fps = stream["avg_frame_rate"]
```
qubist, line 97: Did you test if this works with .mkv files? If not I can give it a shot to make sure. I remember the .mkv file I was working with not having a weird assortment of the fields that the .mp4 had in the probe

missionfloyd: I can still read the framerate from mkv's.
```
pi@pizero:~/tests $ python3 ffprobe.py test.mkv
Framerate: 23.986019736842106
```

robweber (approved):
Just looking at the file finding mechanism as a whole it's really become a tangle of `if` statements. There are a lot of parameters, some conflicting, that define behavior. In some cases they work in unpredictable ways when combined together. 
Once this merged in it might be worth the effort map out a flow and then refactor this to be easier to follow. Maybe introduce a few `class` objects or something to keep track of program state without the tangle of conditionals?

slowmovie.py
```diff
@@ -27,11 +27,10 @@
 # Defaults
 defaultIncrement = 4
 defaultDelay = 120
-defaultContrast = 1.0
```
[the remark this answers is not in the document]
missionfloyd: I figured because 1.0 is no change to contrast (the default) while the others are personal preference. But you're right, they're redundant now that there's a config file.

slowmovie.py
missionfloyd: I've just been running them through a test script to make sure the values are there.
```python
#!/usr/bin/python3
import os, json, sys
import ffmpeg
from fractions import Fraction

if len(sys.argv) > 1:
    videoinfo = ffmpeg.probe(sys.argv[1], select_streams = "v")
    #with open('ffprobe.json', 'w') as f:
    #    json.dump(videoinfo, f)
    #print(f'Frame count: {int(videoinfo["streams"][0]["nb_frames"])}')
    print(f'Framerate: {float(Fraction(videoinfo["streams"][0]["avg_frame_rate"]))}')

```

PR #67, opened 2021-04-19
Subtitle Support

missionfloyd:
out

I've been sitting on this one for a while because I didn't know if It'd be worth adding.

This adds a `--subtitles` argument that, when enabled, Looks for subtitle files matching the video's filename. Supported filetypes are SRT, SSA, and ASS. Any type supported by ffmpeg should work, but those seem to be the most common.

Embedded subtitle tracks *would* work, but it seems to take a lot longer to extract frames, so I've left that out.

Any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ The following options are available:
 ```
 usage: slowmovie.py [-h] [-f FILE] [-R] [-r] [-D DIRECTORY] [-d DELAY]
                     [-i INCREMENT] [-s START] [-c CONTRAST] [-l]
+                    [-o {DEBUG,INFO,WARNING,ERROR,CRITICAL}] [-S | -t]
 
 optional arguments:
   -h, --help            show this help message and exit
@@ -67,7 +68,7 @@ optional arguments:
                         directory order
   -r, --random-frames   choose a random frame every refresh
   -D DIRECTORY, --directory DIRECTORY
-                        videos directory containing available videos to play
+                        directory containing available videos to play
                         (default: Videos)
   -d DELAY, --delay DELAY
                         delay in seconds between screen updates (default: 120)
@@ -80,14 +81,16 @@ optional arguments:
   -l, --loop            loop a single video; otherwise play through the files
                         in the videos directory
   -o {DEBUG,INFO,WARNING,ERROR,CRITICAL}, --loglevel {DEBUG,INFO,WARNING,ERROR,CRITICAL}
-                         minimum importance-level of messages displayed and
-                         saved to the logfile (default: INFO)
+                        minimum importance-level of messages displayed and
+                        saved to the logfile (default: INFO)
+  -S, --subtitles       Display SRT subtitles
+  -t, --timecode        Display video timecode
 
 Args that start with '--' (eg. -f) can also be set in a config file
 (slowmovie.conf). Config file syntax allows: key=value, flag=true,
-stuff=[a,b,c] (for details, see syntax at https://pypi.org/project/ConfigArgParse).
-If an arg is specified in more than one place, then commandline values override
-config file values, which in turn override defaults.
+stuff=[a,b,c] (for details, see syntax at https://goo.gl/R74nmi). If an arg is
+specified in more than one place, then commandline values override config file
+values which override defaults.
 ```
 
 ### Running as a service
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -118,17 +118,7 @@ def video_info(file):
         # Calculate frametime (ms each frame is displayed)
         frameTime = 1000 / fps
 
-        subtitle_file = None
-
-        # Check for a matching subtitle file
-        if args.subtitle_file:
-            name, _ = os.path.splitext(file)
-            for i in glob.glob(name + ".*"):
-                _, ext = os.path.splitext(i)
-                if ext.lower() in subtitle_fileTypes:
-                    subtitle_file = i
-                    logger.debug(f"Found subtitle file '{i}'")
-                    break
+        subtitle_file = find_subtitles(file, probeInfo["streams"])
 
         info = {
             "frame_count": frameCount,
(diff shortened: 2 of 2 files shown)


missionfloyd:
out

`-t` / `--timecode`

See https://ffmpeg.org/ffmpeg-filters.html#drawtext for how this works.

robweber:
I tried to test this out but always get an error when running with `-t`. If I run without it everything works but as soon as I add it I get an error. I'm guessing it's because the call to `overlay_filter()` doesn't do any work until the arg is passed in? Maybe there is a library or something on my system missing. Is it running fine for everyone else? If so it might just be my problem. 

```
Traceback (most recent call last):
  File "./slowmovie.py", line 362, in <module>
    generate_frame(currentVideo, "/dev/shm/frame.bmp", msTimecode)
  File "./slowmovie.py", line 61, in generate_frame
    .run(capture_stdout=True, capture_stderr=True)
  File "/usr/local/lib/python3.7/dist-packages/ffmpeg/_run.py", line 325, in run
    raise Error('ffmpeg', out, err)
ffmpeg._run.Error: ffmpeg error (see stderr output for detail)
```

missionfloyd:
Could you try extracting a frame manually? Here's the filter.
```
-vf "drawtext=text='%{pts\:hms}':fontcolor=white:fontsize=24:x=(w-text_w)/2:y=(h-(text_h*2)):bordercolor=black:borderw=1"
```
Are you using noobs? Some dependency could be missing, or configured weird.

There really needs to be an error handler in generate_frame()

robweber:
I'm using a default install of Raspbian (Raspberry Pi OS) with libraries installed. Running manually I get: 

```
[Parsed_drawtext_0 @ 0x3d13590] No font filename provided
[AVFilterGraph @ 0x3d12df0] Error initializing filter 'drawtext' with args 'text=%{pts\:hms}:fontcolor=white:fontsize=24:x=(w-text_w)/2:y=(h-(text_h*2)):bordercolor=black:borderw=1'

```

Doing some digging it appears ffmpeg needs to be compiled with the option `--enable-libfontconfig` to enable a default fallback font when one isn't specified. I bet that's my issue.  Looking at my configuration options I'm not seeing that.

missionfloyd:
Let's try giving it a font file. Add the parameter `fontfile="/usr/share/fonts/truetype/dejavu/DejaVuSans.ttf"` (that font should be present, run `fc-list` for installed fonts) to line 69.

robweber:
Adding the font file to line 69 made it work. Not sure why my install didn't have that configuration option enabled. More than likely probably an issue with my install specifically. If following the default build instructions others shouldn't have this problem. I can't remember _exactly_ how ffmpeg was installed on this system but I'm wondering if I built it from source. 

Thanks for the help!

qubist:
Before merge I can test this too to make sure it wasn't just @robweber with that issue

missionfloyd:
@qubist Yeah, that'd be helpful, thanks.

slowmovie.py
```diff
@@ -184,6 +213,9 @@ def estimate_runtime(delay, increment, frames, output="guess"):
 parser.add_argument("-c", "--contrast", default=1.0, type=float, help="adjust image contrast (default: %(default)s)")
 parser.add_argument("-l", "--loop", action="store_true", help="loop a single video; otherwise play through the files in the videos directory")
 parser.add_argument("-o", "--loglevel", default="INFO", type=str.upper, choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"], help="minimum importance-level of messages displayed and saved to the logfile (default: %(default)s)")
+group = parser.add_mutually_exclusive_group()
+group.add_argument("-S", "--subtitles", action="store_true", help="Display SRT subtitles")
+group.add_argument("-t", "--timecode", action="store_true", help="Display video timecode")
```
qubist, line 218: Should this be named `textOverlayGroup` or something more specific?

```suggestion
textOverlayGroup = parser.add_mutually_exclusive_group()
textOverlayGroup.add_argument("-S", "--subtitles", action="store_true", help="Display SRT subtitles")
textOverlayGroup.add_argument("-t", "--timecode", action="store_true", help="Display video timecode")
```

qubist, line 218:
```suggestion
group = parser.add_mutually_exclusive_group()
group.add_argument("-S", "--subtitles", action="store_true", help="display SRT subtitles")
group.add_argument("-t", "--timecode", action="store_true", help="display video timecode")
```
for consistency

qubist (changes requested):
Subtitles work fine:
IMG_9732 2

Timecode works fine:
IMG_9733 2

Worked great on the first try, no problems. I most recently built things with the Install script, and this is on a NOOBS-derived Pi 4.

I left some minor suggestions. Also README.md needs to be updated with the new options.

TomWhitwell:
> Subtitles work fine:
> IMG_9732 2
> 
> Timecode works fine:
> IMG_9733 2
> 
> Worked great on the first try, no problems. I most recently built things with the Install script, and this is on a NOOBS-derived Pi 4.
> 
> I left some minor suggestions. Also README.md needs to be updated with the new options.

same film playing here!  
image

robweber:
This looks really cool. I think with normal installs this will work great. Assume my issues were a fluke!

README.md
```diff
@@ -80,14 +81,16 @@ optional arguments:
...
+                        saved to the logfile (default: INFO)
+  -S, --subtitles       Display SRT subtitles
+  -t, --timecode        Display video timecode
 
 Args that start with '--' (eg. -f) can also be set in a config file
 (slowmovie.conf). Config file syntax allows: key=value, flag=true,
-stuff=[a,b,c] (for details, see syntax at https://pypi.org/project/ConfigArgParse).
-If an arg is specified in more than one place, then commandline values override
-config file values, which in turn override defaults.
+stuff=[a,b,c] (for details, see syntax at https://goo.gl/R74nmi). If an arg is
```
qubist, line 91: Should this stay a shortlink? I remember @missionfloyd changing it [in a past commit](3e1891ebcf2d876cdaac0350e3c1ce65dc703e8c) and I agree that the longer link is nicer because it shows where it's leading. But the actual output of `slowmovie.py -h` *does* have the shortlink in it. Thoughts one way or the other?

qubist (approved):
Looks wonderful

README.md
missionfloyd: I forgot about that. I just copied the entire output.

PR #64, opened 2021-04-15
Multiple Displays

robweber:
This is my second draft at this. The first one was way too cumbersome and didn't fit with the focus of this project. This one takes more of a library approach and involves minimal changes to the codebase here. 

## Goal 

Provide a way to allow multiple display types to be used with `slowmovie.py` with the most minimal impact to the developing codebase here. The solution should be user friendly and not require end users to muck about changing python files. 

## Solution 

I've spun off a [separate repo](https://github.com/robweber/vsmp-epd) specifically for abstracting EPD drivers for "vsmp" (very slow movie player) projects. Right now it will handle a mock display (for testing without hardware) and any Waveshare devices in the [Waveshare e-Ink repo](https://github.com/waveshare/e-Paper). Within the next few days I hope to add Inky displays as well, just waiting for my Inky hardware to arrive. 

`vsmp-epd` is published to PyPi and is available as a dependency. This makes it easy to integrate with this project. You can see from the changes to `slowmovie.py` that it's almost a 1:1 replacement in terms of interacting with the library. The main 7.5in display from the instructions is the default but using the `-e` argument you can specify any compatible display. 

## Pros/Cons

The big pro here is this keeps all the E-ink abstraction code separate from this project. The goal here is to provide the media player, not drag along the interaction code for the displays. It also allows for specifying different display types pretty easily for the end user. 

The major con is mainly Waveshare itself. Because it's not available as a `pip` install it will still be a manual build per the install instructions. I can include the Inky libraries as dependencies with `vsmp-epd` but Waveshare will need to have it's own build instructions (unless you're not using that type of display). 

__Fun Note__: I've been running this abstraction code for 2 days now with no issues on my home display.

PR #71, opened 2021-05-02
Default Config File

robweber:
As I was doing some testing it occurred to me that custom changes to `slowmovie.conf` are either lost ,or you get a conflict, when doing updates to the repo directory with `git pull` or the install script. This moves the default configuration file to the Install directory and only copies it when missing as part of the install process. Adding it to `.gitignore` assures it isn't messed with on accident. 

This also fixes the issue [pointed out](https://github.com/TomWhitwell/SlowMovie/pull/64#discussion_r624146244) by @missionfloyd in #64 where it is a good idea to have some sane defaults a user can revert to in the event they miss with, or even delete, the default config file.


missionfloyd (approved):
Good idea. The contents of the config file could probably use a once-over. it's just how I happened to leave it. It could be populated with default values (possibly commented out), and descriptions.

PR #78, opened 2021-05-29
Log argparse errors

missionfloyd:
As suggested by @jschrempp, this enables logging for argparse errors.
```console
pi@pizero:~/SlowMovie $ ./slowmovie.py -f null.mp4
argument -f/--file: File 'null.mp4' does not exist
```
```
[2021-05-29 14:45:03,231] ERROR   : argument -f/--file: File 'null.mp4' does not exist
```

### DIFF
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -201,7 +201,25 @@ def find_subtitles(file):
     return None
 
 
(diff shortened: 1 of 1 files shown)


robweber:
Makes sense to me. Weird you have to extend the parser class to get this kind of functionality. Will be nice to see this info in the log file when running as a service.

Issue #70, opened 2021-04-28
Install script upgrade doesn't update Waveshare library


qubist:
**Describe the problem**
> Simply running the script to upgrade did not appear to update the waveshare library.

> Since [Waveshare] don't [use version numbers] the version always comes in as 0.0.0. When you re-run the script it sees the same version, assumes the code is the same, and won't rebuild.

> maybe something for us to address as part of the update script. Maybe always uninstall/reinstall waveshare? Takes up time but might save headaches.

— @robweber and @grooverkm in #69

**Steps to Reproduce**
1. Have everything installed with outdated version of Waveshare library
2. Run Install script and select upgrade

Waveshare is not updated to the newest version

qubist:
I think always uninstalling and reinstalling Waveshare is fine. How hard would it be to make a PR over there for them to use version numbers also?

robweber:
Not too hard really, they would just need to include a version number as part of their [setup.py ](https://github.com/waveshare/e-Paper/blob/master/RaspberryPi_JetsonNano/python/setup.py)file. The harder sell would probably be getting them to actually use it and increment it when things change. That would be the _correct_ way to fix this.

PR #80, opened 2021-06-01
Python Libs Update

robweber:
When running the `install.sh` file Python lib dependencies are not updated. It appears as though when installing from a git URL using pip it will not honor version updates in a setup.py or setup.cfg file. 

The fix for this is to use the `-U` flag to force updates of all dependencies to the latest version. This works great but one question before merging I have is if we want to start specifying specific versions within the `requirements.txt` file? Our requirements for these libraries are pretty minimal so it's unlikely updates will break things; however by specifying known working versions we are covering our bases as well. For Git URLs you can specify a tagged revision so for libs like `omni-epd` we can add `@v0.2.3` as part of the URL string.

Of course the downside to this is that we'll need a lot more hand-holding to the `requiements.txt` file as versions of the libraries change. Any thoughts on this?

### DIFF
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -23,7 +23,7 @@ function install_python_packages(){
 
 
   sudo pip3 install setuptools -U
-  sudo pip3 install -r $LOCAL_DIR/Install/requirements.txt
+  sudo pip3 install -r $LOCAL_DIR/Install/requirements.txt -U
 }
 
 function setup_hardware(){
--- a/Install/requirements.txt
+++ b/Install/requirements.txt
@@ -1,4 +1,5 @@
 ffmpeg-python==0.2.0
 Pillow==8.2.0
 ConfigArgParse==1.4.1
+git+https://github.com/waveshare/e-Paper.git#egg=waveshare-epd&subdirectory=RaspberryPi_JetsonNano/python
 git+https://github.com/robweber/omni-epd.git@v0.2.3#egg=omni-epd
--- a/Install/requirements.txt
+++ b/Install/requirements.txt
@@ -1,5 +1,4 @@
 ffmpeg-python
 pillow
 ConfigArgParse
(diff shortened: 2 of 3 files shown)


missionfloyd:
Makes sense, we know it works with at least the current versions.

Should waveshare-epd stay in requirements.txt? We're not using it directly anymore.

robweber:
I can pull out the waveshare libs since `omni-epd` will pull them in. Either way they'll get installed so shouldn't be a big deal. I'll update the requirements file with the most current versions numbers as well of the libs we do have in there. 

I also believe this will satisfy #70 when it's merged in. The `-U` option should always pull in Git URLs so that fix that issue too.

robweber:
One weird quirk to removing the waveshare git URL from the `requirements.txt` file has surfaced. It appears that when `pip install -U` is used with the `-r` option it will download and rebuild dependencies from URLs, ie the git sourced libraries. This does not happen when git URLs are sourced in a `setup.py` or `setup.cfg` file as they are done in the `omni-epd` library. 

I even went so far in testing as to specify specific commits on the waveshare repo and these are ignored as well, even using `-U`. Documentation I've found states that `pip` will not respect git tags or commit hashes as part of a versioning scheme. This seemingly makes it impossible to ensure the waveshare libs are updated since waveshare doesn't submit to PyPi or use versioning on their repo. 

Seems the only surefire way to make sure it always gets updated is to include it in the `requirements.txt` file.

missionfloyd:
Might as well leave it then.

robweber:
Added the waveshare repo back in to `requirements.txt`. Tested it and when using `pip3 install -U` it will download and rebuild the library each time now. I did use the tagged version of the `omni-epd` repo as part of the URL so we can specify specific versions of that like we can with PyPi packages. It will rebuild each time as well since it's a repo URL but it will go to that specific tag.

robweber:
Going to merge this in - any issues with documentation can be cleaned up later if needed. This way we can get the library re-building during updates.

PR #81, opened 2021-06-01
Run pip without sudo

missionfloyd:
I just tried this on a freshly-imaged card, works fine this way. Pip isn't really supposed to be run as root anyway. Everything goes in `~/.local/lib` instead. It does produce warnings like this:
```
The script omni-epd-test is installed in '/home/pi/.local/bin' which is not on PATH.
Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
```
...but this is already handled in `.profile`.
```shell
# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/.local/bin" ] ; then
    PATH="$HOME/.local/bin:$PATH"
fi
```

I've also removed the "update setuptools" step. That doesn't seem to be necessary anymore.

### DIFF
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -15,15 +15,11 @@ SERVICE_FILE=slowmovie.service
 
 function install_linux_packages(){
   sudo apt-get update
-
   sudo apt-get install -y ffmpeg git python3-pip
 }
 
 function install_python_packages(){
-
-
-  sudo pip3 install setuptools -U
-  sudo pip3 install -r $LOCAL_DIR/Install/requirements.txt
+  pip3 install -r $LOCAL_DIR/Install/requirements.txt
 }
 
 function setup_hardware(){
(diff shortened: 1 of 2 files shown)


robweber:
That makes sense to me. I did some playing around on my install and it appears as though if the library already exists in  `/usr/local/lib/python...` it will use that one but any new packages are then downloaded to the local profile. Makes sense since you'd have read access to that directory as a user but need `sudo` to write new. 

I don't have a new install to test it but running as a service later won't run in to issues right? That will run with `sudo` but since we're specifying the [user in the service file](https://github.com/TomWhitwell/SlowMovie/blob/main/slowmovie.service#L5) it should still find the locally installed packages for that user right?

missionfloyd:
Yeah, running it as a service still works,

robweber:
This makes sense to me, no reason to run stuff with elevated privileges unless we need to.

PR #91, opened 2021-08-01
Add support for brightness

ceicke:
Add support for changing the brightness of a picture in the post-processing step. Useful for very dark movies.


missionfloyd:
Already supported. See https://github.com/TomWhitwell/SlowMovie#e-ink-display-customization

ceicke:
But I guess for code cleanup, then the `contrast` switch and the corresponding code can be removed.

PR #101, opened 2021-10-27
clear display on exit

tkameroski:
Clear display when exit handler is triggered.

### DIFF
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -35,9 +35,10 @@ os.chdir(os.path.dirname(os.path.realpath(__file__)))
 
 # Handle when the program is killed and exit gracefully
 def exithandler(signum, frame):
-    epd.prepare()
-    epd.clear()
     logger.info("Exiting Program")
+    if args.clear:
+        epd.prepare()
+        epd.clear()
     try:
         epd.close()
     finally:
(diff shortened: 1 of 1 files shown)


robweber:
Is clearing the screen on exit something everyone wants? Sometimes I'll kill the program to leave a particular image up. The bonus to e-ink is that it will keep the image even without power. Can you define a use case where this would be good practice?

tkameroski:
I was thinking that a common use for a shutdown would be for storage of the player and that clearing it would prevent damage of ghosting if it were shutdown for a long time.

missionfloyd:
I don't think that's really an issue with e-ink screens.

I suppose an option could be added.

robweber:
I think that makes sense. Could you update this to add it as a passed in arg? That way those than want it can use it but we won't force it by default.

slowmovie.py
```diff
@@ -234,6 +237,7 @@ def error(self, message):
 parser.add_argument("-l", "--loop", action="store_true", help="loop a single video; otherwise play through the files in the videos directory")
 parser.add_argument("-e", "--epd", help="the name of the display device driver to use")
 parser.add_argument("-o", "--loglevel", default="INFO", type=str.upper, choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"], help="minimum importance-level of messages displayed and saved to the logfile (default: %(default)s)")
+parser.add_argument("-C", "--clear", action="store_true", help="clears displays when exithandler is triggered")
```
missionfloyd, line 240:
```suggestion
parser.add_argument("-C", "--clear", action="store_true", help="clear display on exit")
```
A little clearer, I think.

Issue #111, opened 2022-01-16
Stops working in Random File Random Movie Mode


ElDuderino79:
**Describe the problem**
Random File & Random Movie = True, Update Interval 60, several movies, slow movie service stops.

Hi there,
a fantastic project! Love it! And would like to use it with several movies and random file random frame to have an always changing picture frame. Unfortunately, it seems to stop working shortly after reboot.

My config file looks like this:

> random-file = True
> random-frames = True
> delay = 120
> increment = 8
> contrast = 2.0
> epd = waveshare_epd.epd7in5_V2

Unfortunately, after some frames, it seems to stop working. Log files looks as follows:

> [2022-01-16 13:26:08,546] INFO    : slowmovie : Update interval: 120
> [2022-01-16 13:26:12,000] INFO    : slowmovie : Playing 'Movie_A.mp4'
> [2022-01-16 13:26:12,000] INFO    : slowmovie : Video info: 196940 frames, 23.976fps, duration: 02:16:54

Display the status of the SlowMovie service:

> ● slowmovie.service - Slow Movie Player Service
>      Loaded: loaded (/etc/systemd/system/slowmovie.service; enabled; vendor preset: enabled)
>      Active: failed (Result: exit-code) since Sun 2022-01-16 13:26:30 CET; 6min ago
>     Process: 440 ExecStart=/usr/bin/python3 /home/pi/SlowMovie/slowmovie.py (code=exited, status=1/FAILURE)
>    Main PID: 440 (code=exited, status=1/FAILURE)
>         CPU: 9.086s
> 
> Jan 16 13:26:29 SlowMovieFrame python3[440]:   File "/home/pi/SlowMovie/slowmovie.py", line 404, in <module>
> Jan 16 13:26:30 SlowMovieFrame python3[440]:     videoInfo = video_info(currentVideo)
> Jan 16 13:26:30 SlowMovieFrame python3[440]:   File "/home/pi/SlowMovie/slowmovie.py", line 108, in video_info
> Jan 16 13:26:30 SlowMovieFrame python3[440]:     probeInfo = ffmpeg.probe(file, select_streams="v")
> Jan 16 13:26:30 SlowMovieFrame python3[440]:   File "/home/pi/.local/lib/python3.9/site-packages/ffmpeg/_probe.py", line 23, in probe
> Jan 16 13:26:30 SlowMovieFrame python3[440]:     raise Error('ffprobe', out, err)
[...]

PR #114, opened 2022-02-01
Filter out hidden files

robweber:
This will solve the problem found in #111 where hidden files that weren't actually video files were being included in the file listing. `os.listdir()` returns hidden system files that start with a dot (`.`). This fix will filter them out at the time the file list is being generated. In the issue referenced above files such as `._test.mp4` were being included. I imagine this happens when entire directories are copied from other computers (Mac, Windows) that have generated things like thumbnails that include the full file name. 

I tried a few different ways of doing this, including using the `glob` package; but this seemed the most efficient. glob returned the absolute path to the files instead of just the filename, which would have created a bunch of downstream problems as we're expecting just the filename in other places of the code. I also didn't want to iterate through the list twice since we're doing the video file ext filter anyway. Doing them at the same time seemed the easiest and got rid of the hidden files.

### DIFF
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -97,7 +97,7 @@ def check_dir(value):
 
 def supported_filetype(file):
     _, ext = os.path.splitext(file)
-    return ext.lower() in fileTypes
+    return ext.lower() in fileTypes and not file.startswith('.')
 
 
 # Get framerate, frame count, duration, and frame-time of video via FFmpeg probe


qubist (approved):
This looks good. I can't think of any scenario we'd want to have the code interact with a hidden file, even if it _is_ a valid movie file. It's intuitive that hidden movies won't be played. And this will fix the issue.

Issue #115, opened 2022-02-21
Runtime estimation problem


yippieyaray:
Hi Tom!

There is a problem estimating the runtime (estimate_runtime) if the number of remaining frames is equal or below the increment.

The reason is that the for-loop in line 182 uses a wrong result variable named outputGuess instead of output:

for length, outputGuess in [(days, "d"), (hours, "h"), (minutes, "m"), (seconds, "s")]:

If the calculated seconds is 1 or below, the loop runs to the end and the variable output keeps the value "guess".

The subsequent if-block steps to line 196 (call stack see below)

Changing the line 182 to the following line fixes the error:
for length, output in [(days, "d"), (hours, "h"), (minutes, "m"), (seconds, "s")]

Regards,
Ray 

call stack:

Feb 21 11:48:00 raspberrypi python3[1364]: Traceback (most recent call last):                                            Feb 21 11:48:00 raspberrypi python3[1364]:   File "/home/pi/SlowMovie/slowmovie.py", line 378, in <module>               Feb 21 11:48:00 raspberrypi python3[1364]:     logger.info(f"This video will take {estimate_runtime(args.delay, args.inc>Feb 21 11:48:00 raspberrypi python3[1364]:   File "/home/pi/SlowMovie/slowmovie.py", line 196, in estimate_runtime       Feb 21 11:48:00 raspberrypi python3[1364]:     raise ValueError                                                          Feb 21 11:48:00 raspberrypi python3[1364]: ValueError

qubist:
Hi @yippieyaray, and thanks for your bug report. Would you please create a PR for this fix so we can test and merge it?

weow147:
interesting... updated.  I'll keep an eye on it.

qubist:
FWIF, I propose the following change to the `estimate_runtime` function, which fixes the issue you bring up, but also preserves code readability a bit better IMO than changing `outputGuess`

```python3
    if output == "guess":
        # Choose the biggest units that result in a quantity greater than 1
        # If no unit results in a quantity greater than 1, choose seconds
        for length, unit in [(days, "d"), (hours, "h"), (minutes, "m"), (seconds, "s")]:
            if length > 1 or unit == "s":
                return estimate_runtime(delay, increment, frames, unit)
```

(Note I renamed `outputGuess` to `unit` cause it makes more sense!)

Another alternative is as follows:

```python3
    if output == "guess":
        # Choose the biggest units that result in a quantity greater than 1
[...]

PR #117, opened 2022-02-22
Estimate runtime refactoring

yippieyaray:
This is my refactoring of the estimate_runtime function to fix "Runtime estimation problem #115 (https://github.com/TomWhitwell/SlowMovie/issues/115)"
It's not my quick fix from yesterday. I remove all unnecessary loops and recursions as well.

### DIFF
--- a/slowmovie.py
+++ b/slowmovie.py
@@ -165,36 +165,27 @@ def get_random_video(viddir):
 
 
 # Calculate how long it'll take to play a video.
-# output value: "d[ay]", "h[our]", "m[inute]", "s[econd]", "all"; omit for an automatic guess
-def estimate_runtime(delay, increment, frames, output="guess"):
-    # Recurse to generate all estimates in one string
-    if output == "all":
-        return f"{estimate_runtime(delay, increment, frames, 's')} / {estimate_runtime(delay, increment, frames, 'm')} / {estimate_runtime(delay, increment, frames, 'h')} / {estimate_runtime(delay, increment, frames, 'd')}"
-
+# output value: "d[ay]", "h[our]", "m[inute]", "s[econd]", false; omit for an automatic guess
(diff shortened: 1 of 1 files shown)


robweber:
The changes make sense to me. I ran the new function in it's own file with a variety of different increments and frame values without issue. The only spot I can find this function is called is when the [remaining time is logged](https://github.com/TomWhitwell/SlowMovie/blob/main/slowmovie.py#L378). For it's intended purpose I think making it as lightweight as possible is a good change. 

@qubist - I think you may have originally added this function. Any thoughts?

qubist:
If merged, this closes #115

slowmovie.py
```diff
@@ -165,35 +165,27 @@ def get_random_video(viddir):
 
 
 # Calculate how long it'll take to play a video.
-# output value: "d[ay]", "h[our]", "m[inute]", "s[econd]", "all"; omit for an automatic guess
-def estimate_runtime(delay, increment, frames, output="guess"):
-    # Recurse to generate all estimates in one string
-    if output == "all":
-        return f"{estimate_runtime(delay, increment, frames, 's')} / {estimate_runtime(delay, increment, frames, 'm')} / {estimate_runtime(delay, increment, frames, 'h')} / {estimate_runtime(delay, increment, frames, 'd')}"
-
+# output value: "d[ay]", "h[our]", "m[inute]", "s[econd]", false; omit for an automatic guess
```
qubist, line 168:
```suggestion

```

qubist (changes requested):
This is a very elegant refactor of the function! Thank you for your contribution, @yippieyaray. I'm ready to approve changes after the comment line I mentioned is removed.

qubist:
Thanks for moving forward with this, @missionfloyd !

qubist (approved):
I'm ready for this to be merged

PR #136, opened 2022-09-20
Dynamic Install Path

robweber:
With newer versions of the Raspberry Pi OS you can't count on the default `pi` user being the path for the home folder anymore. This PR attempts to modify the install procedure and the install script to insert the current user name in places where a full path is needed. For the `install.sh` script this is pretty simple. I just replaced any hardcoded paths from `/home/pi` to `/home/$USER`. 

The one piece that was a little tricky was the `slowmovie.service` file. This needed to be updated on the fly rather than a static file with hardcoded paths. I utilized the [envsubst](https://manpages.debian.org/unstable/gettext-base/envsubst.1.en.html) shell function to generate the service file from a template and then move that file to the correct location. 

### Limitations

This is a 1:1 replacement for the current install procedure. It doesn't try and account for more dynamic path options like someone having a completely different path for the SlowMovie install other than as a subdirectory in the user's home folder (`/home/$USER/SlowMovie`). That could probably be added based on these modifications but that wasn't the intent here.

### DIFF
--- a/slowmovie.service.template
+++ b/slowmovie.service.template
@@ -3,8 +3,8 @@ Description=Slow Movie Player Service
 
 [Service]
 User=${USER}
-WorkingDirectory=/home/${USER}/SlowMovie
-ExecStart=/usr/bin/python3 /home/${USER}/SlowMovie/slowmovie.py
+WorkingDirectory=${HOME}/SlowMovie
+ExecStart=/usr/bin/python3 ${HOME}/SlowMovie/slowmovie.py
 StandardOutput=null
 StandardError=journal
 
--- a/slowmovie.service.template
+++ b/slowmovie.service.template
@@ -2,9 +2,9 @@
 Description=Slow Movie Player Service
 
 [Service]
-User=pi
-WorkingDirectory=/home/pi/SlowMovie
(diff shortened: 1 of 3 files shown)


missionfloyd:
`/home/$USER` can be shortened to `$HOME` or `~`

robweber:
Good thought - I was going for simplest solution (literally did a find replace) but using $HOME makes more sense.

slowmovie.service.template
```diff
@@ -2,9 +2,9 @@
 Description=Slow Movie Player Service
 
 [Service]
-User=pi
-WorkingDirectory=/home/pi/SlowMovie
-ExecStart=/usr/bin/python3 /home/pi/SlowMovie/slowmovie.py
+User=${USER}
+WorkingDirectory=/home/${USER}/SlowMovie
+ExecStart=/usr/bin/python3 /home/${USER}/SlowMovie/slowmovie.py
```
missionfloyd, line 7:
```suggestion
WorkingDirectory=${HOME}/SlowMovie
ExecStart=/usr/bin/python3 ${HOME}/SlowMovie/slowmovie.py
```
Missed one.

Issue #141, opened 2022-10-23
Selecting NO to service install exits script


robweber:
**Describe the problem**
When doing a clean install using the `install.sh` script the script stops functioning if "no" is selecting during the _run as a service?_ prompt. Instead of continuing properly and not installing the service the script just halts and exits. Selecting "yes" continues as expected. 

**Steps to Reproduce**
Do a clean install of SlowMovie using the current install script. When prompted to install the service select no. Make sure SlowMovie is completely uninstalled if trying from an existing system (uninstall service, delete git folder, etc). 

**Platform Information** 

Was tested on an Rpi from a clean install.

robweber:
Wanted to add this was tested on both Raspberry Pi OS 10 and 11 with the same results.

Issue #140, opened 2022-10-22
Issue with slowmovie.service


MasonPi:
When doing a fresh install from the automatic setup, I'm getting an error with the slowmovie.service 

cp: cannot stat 'slowmovie.service': No such file or directory
Failed to enable unit: Unit file slowmovie.service does not exist.

Is there any fix for this?

robweber:
Looks like you found an error from when this was recently updated. The template file code from [here](https://github.com/TomWhitwell/SlowMovie/blob/main/Install/install.sh#L93) should also be updated [here](https://github.com/TomWhitwell/SlowMovie/blob/main/Install/install.sh#L111). 

Just "thinking out loud" the best solution would be to make a function for this so it isn't the same code scattered in two spots for future updates. I'll try and update the install file by the end of the weekend (unless someone else beats me to it!). In the meantime you should be able to run this via the command line from the SlowMovie directory to get the service working. 

```
envsubst <slowmovie.service.template > slowmovie.service
sudo mv slowmovie.service /etc/systemd/system
sudo systemctl daemon-reload
sudo systemctl enable slowmovie
```

PR #142, opened 2022-10-23
Service install Fixes

robweber:
These should fix both #140  and #141. 

For the first one the service install process was not working properly as it wasn't generating the service file from the template as expected.  Some reconfiguring of the service install process was done to fix this. I added a `copy_service_file` function to keep all this code in the same spot. The `install_slowmovie` function now calls `install_service` directly when it notices the service is installed during an upgrade. `install_service` now branches to either do a new service install or an upgraded service (if necessary). 

For the second bug I actually found this while trying to fix the first one. When selecting "no" to the service install question the entire script was just closing out instead of continuing properly. I played around with `whiptail` a bunch but it just didn't seem to be returning the variable properly no matter what I did. Storing the result directly using either `$?` or assigning it as part of the call all resulted in the script crashing. Wrapping everything in an `if` statement was the only way I could get it to function. Now the `INSTALL_SERVICE` variable is set based on the yes/no action instead of storing the output of whiptail directly. This worked when testing in all scenarios (selecting yes or no during clean install).

### DIFF
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -46,6 +46,11 @@ function service_installed(){
   fi
 }
 
+function copy_service_file(){
+  sudo mv $SERVICE_FILE $SERVICE_DIR
+  sudo systemctl daemon-reload
+}
+
 function install_slowmovie(){
   FIRST_TIME=1  # if this is a first time install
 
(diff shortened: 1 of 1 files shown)


robweber:
Tested on two different systems (Rpi OS 10 and 11) - all seems to be working. Going to merge it in.

PR #145, opened 2023-03-14
Allow For Waveshare Driver Rollback

robweber:
This is an attempt to give end-users more information and options regarding the known streaking issues with some Waveshare displays (notably the 7.5in V2) as described in several issues. 

This PR adds a disclaimer to the README and adds an additional option to the automated installer that will rollback the driver to a known working version. It is worth noting that subsequent runs of the install script will attempt to upgrade the Waveshare drivers since Waveshare doesn't use version numbers as part of their install. Users that need the rollback can re-run the rollback option on update. This allows for the option of always reverting back to the most up to date drivers if needed via the "update" method.

### DIFF
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -23,6 +23,14 @@ function install_python_packages(){
   pip3 install -r $LOCAL_DIR/Install/requirements.txt -U
 }
 
+function rollback_waveshare(){
+  # uninstall waveshare lib (installed with omni-epd)
+  pip3 uninstall waveshare-epd
+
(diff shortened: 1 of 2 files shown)


robweber:
Was just about to merge this in and noticed [this PR](https://github.com/waveshare/e-Paper/pull/289). Going to wait and see if it gets merged in, that would make all this unnecessary.

robweber:
Going to close this out. The changes from @missionfloyd have been merged in to the Waveshare branch. I did a new install from the Waveshare repo and the streaking is gone now. No need for these fixes.

Issue #150, opened 2023-10-12
Install broken due to python changes in Raspbian Bookworm


JonCellini:
**Describe the problem**
Due to changes in how python functions in Raspbian Bookworm the default install script fails to install the python dependencies due to a change in how that is handled by default.

It gives the following error:

**error: externally-managed-environment**

The script continues but SlowMovie will fail to run as it's dependencies are not properly met.

```
Cloning into '/home/pi/SlowMovie'...
remote: Enumerating objects: 1479, done.
remote: Counting objects: 100% (543/543), done.
remote: Compressing objects: 100% (168/168), done.
remote: Total 1479 (delta 474), reused 375 (delta 375), pack-reused 936
Receiving objects: 100% (1479/1479), 5.55 MiB | 7.96 MiB/s, done.
Resolving deltas: 100% (832/832), done.
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    For more information visit http://rptl.io/venv

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
SlowMovie install/update complete. To test, run 'python3 /home/pi/SlowMovie/slowmovie.py'
Created symlink /etc/systemd/system/multi-user.target.wants/slowmovie.service → /etc/systemd/system/slowmovie.service.
SlowMovie service installed! Use sudo systemctl start slowmovie to test 
```

**Steps to Reproduce**
Install a fresh Raspbian Bookworm image. Run the installer script.

**Platform Information** 
Raspberry Pi 3 model A+
Raspbian Bookworm 2023-10-10 (reproducible on both armhf & arm64)
Python 3.11.2

robweber:
Doing a bit of research on this it seems this is a direct result of [PEP 668](https://peps.python.org/pep-0668/). A good (and short) explanation of it [is here](https://github.com/python/cpython/issues/102134#issuecomment-1445428402). 

[...]

PR #151, opened 2023-10-30
add code to handle python venv

JonCellini:
Added code to handle the creation and use of a python venv for SlowMovie.

Addresses issue #150

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,5 @@ Videos/*
 !Videos/test.mp4
 slowmovie.log
 slowmovie.conf
+# python venv
+.venv
--- a/.gitignore
+++ b/.gitignore
@@ -76,4 +76,4 @@ Videos/*
 slowmovie.log
 slowmovie.conf
 # python venv
-.venv
+.SlowMovie
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -106,7 +106,8 @@ function install_slowmovie(){
     install_service
   fi
 
-  echo -e "SlowMovie install/update complete. To test, run '${YELLOW}python3 ${LOCAL_DIR}/slowmovie.py${RESET}'"
+  # the venv will not be active in the current users shell until they log out/in  
+  echo -e "SlowMovie install/update complete. To test, run '${YELLOW}source $LOCAL_DIR/.SlowMovie/bin/activate; python3 ${LOCAL_DIR}/slowmovie.py${RESET}'"
 
   return $FIRST_TIME
 }
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -22,6 +22,9 @@ function install_linux_packages(){
 function create_python_venv(){
   python3 -m venv --system-site-packages $LOCAL_DIR/.venv # numpty requires system-site-packages
   source $LOCAL_DIR/.venv/bin/activate #activate our venv
+  # add venv to .profile so it is sourced for the users login shell
+  echo " # activate the SlowMovie python venv"
+  echo "source $LOCAL_DIR/.venv/bin/activate" >> ~/.profile
 }
 
 function install_python_packages(){
@@ -94,7 +97,6 @@ function install_slowmovie(){
     create_python_venv
     # install any needed python packages
     install_python_packages
-
   fi
 
   cd $LOCAL_DIR
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -25,7 +25,8 @@ function create_python_venv(){
 }
 
 function install_python_packages(){
-  pip3 install -r $LOCAL_DIR/Install/requirements.txt -U
+  #fix for IT8951 install https://github.com/GregDMeyer/IT8951/issues/58
+  USE_CYTHON=1 pip3 install -r $LOCAL_DIR/Install/requirements.txt -U
 }
 
 function setup_hardware(){
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -23,8 +23,11 @@ function create_python_venv(){
   python3 -m venv --system-site-packages $LOCAL_DIR/.SlowMovie # numpty requires system-site-packages
   source $LOCAL_DIR/.SlowMovie/bin/activate #activate our venv
   # add venv to .profile so it is sourced for the users login shell
-  echo " # activate the SlowMovie python venv"
-  echo "source $LOCAL_DIR/.SlowMovie/bin/activate" >> ~/.profile
+  if ! grep -q "source $LOCAL_DIR/.SlowMovie/bin/activate" "$HOME/.profile" ; then
+    echo >> "$HOME/.profile"
+    echo "# activate the SlowMovie python venv" >> "$HOME/.profile"
+    echo "source $LOCAL_DIR/.SlowMovie/bin/activate" >> "$HOME/.profile" 
+  fi
 }
 
 function install_python_packages(){
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -20,11 +20,11 @@ function install_linux_packages(){
 }
 
 function create_python_venv(){
-  python3 -m venv --system-site-packages $LOCAL_DIR/.venv # numpty requires system-site-packages
-  source $LOCAL_DIR/.venv/bin/activate #activate our venv
+  python3 -m venv --system-site-packages $LOCAL_DIR/.SlowMovie # numpty requires system-site-packages
+  source $LOCAL_DIR/.SlowMovie/bin/activate #activate our venv
   # add venv to .profile so it is sourced for the users login shell
   echo " # activate the SlowMovie python venv"
-  echo "source $LOCAL_DIR/.venv/bin/activate" >> ~/.profile
+  echo "source $LOCAL_DIR/.SlowMovie/bin/activate" >> ~/.profile
 }
 
 function install_python_packages(){
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -26,6 +26,7 @@ function create_python_venv(){
 
 function install_python_packages(){
   #fix for IT8951 install https://github.com/GregDMeyer/IT8951/issues/58
+  pip3 install cython #install cython into our venv first so we can use it to build the IT8951 package
   USE_CYTHON=1 pip3 install -r $LOCAL_DIR/Install/requirements.txt -U
 }
 
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -16,7 +16,12 @@ SERVICE_FILE_TEMPLATE=slowmovie.service.template
 
 function install_linux_packages(){
   sudo apt-get update
-  sudo apt-get install -y ffmpeg git python3-pip libatlas-base-dev
+  sudo apt-get install -y ffmpeg git python3-pip python3-venv libatlas-base-dev
+}
+
+function create_python_venv(){
+  python3 -m venv $LOCAL_DIR/.venv
+  source $LOCAL_DIR/.venv/bin/activate #activate our venv
 }
 
 function install_python_packages(){
@@ -85,6 +90,8 @@ function install_slowmovie(){
   fi
 
   if [ "$SKIP_DEPS" = false ]; then
+    # create the python venv
+    create_python_venv
     # install any needed python packages
     install_python_packages
 
--- a/Install/install.sh
+++ b/Install/install.sh
@@ -20,11 +20,13 @@ function install_linux_packages(){
 }
 
 function create_python_venv(){
-  python3 -m venv $LOCAL_DIR/.venv
+  python3 -m venv --system-site-packages $LOCAL_DIR/.venv # numpty requires system-site-packages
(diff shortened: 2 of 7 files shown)


robweber:
Thanks for tackling this. Have you fully tested it on Bullseye or Bookworm? I'll clone it to my local instance tonight and give it a run. My local SlowMovie is still on Bullseye so I'll at least be able to test backwards compatibility.

JonCellini:
I'll test it on my local RPI here in a bit and post back. My player is also on Bullseye at the moment so shouldn't be a problem to test on both Bullseye and Bookworm.

JonCellini:
It works fine on Bullseye arm64 :)

The venv code I wrote works as expected on Bookworm arm64; however, SlowMovie fails to install due to an unrelated issue with the IT8951 package.

The failure output is here:

`  Running setup.py install for RPi.GPIO ... done
  DEPRECATION: IT8951 is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
  Running setup.py install for IT8951 ... error
  error: subprocess-exited-with-error
  
  × Running setup.py install for IT8951 did not run successfully.
  │ exit code: 1
  ╰─> [54 lines of output]
      running install
      /home/pi/SlowMovie/.venv/lib/python3.11/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
        warnings.warn(
      running build
      running build_py
      creating build
      creating build/lib.linux-aarch64-cpython-311
      creating build/lib.linux-aarch64-cpython-311/IT8951
      copying IT8951/interface.py -> build/lib.linux-aarch64-cpython-311/IT8951
      copying IT8951/constants.py -> build/lib.linux-aarch64-cpython-311/IT8951
      copying IT8951/display.py -> build/lib.linux-aarch64-cpython-311/IT8951
      copying IT8951/__init__.py -> build/lib.linux-aarch64-cpython-311/IT8951
      running build_ext
      building 'IT8951.spi' extension
      creating build/temp.linux-aarch64-cpython-311
      creating build/temp.linux-aarch64-cpython-311/IT8951
      aarch64-linux-gnu-gcc -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -fPIC -I/home/pi/SlowMovie/.venv/include -I/usr/include/python3.11 -c IT8951/spi.c -o build/temp.linux-aarch64-cpython-311/IT8951/spi.o
      IT8951/spi.c: In function ‘__Pyx_TraceSetupAndCall’:
      IT8951/spi.c:22082:37: error: invalid use of incomplete typedef ‘PyFrameObject’ {aka ‘struct _frame’}
      22082 |         if (CYTHON_TRACE && (*frame)->f_trace == NULL) {
            |                                     ^~
      IT8951/spi.c:22084:21: error: invalid use of incomplete typedef ‘PyFrameObject’ {aka ‘struct _frame’}
      22084 |             (*frame)->f_trace = Py_None;
            |                     ^~
      IT8951/spi.c:438:62: error: invalid use of incomplete typedef ‘PyFrameObject’ {aka ‘struct _frame’}
        438 |   #define __Pyx_PyFrame_SetLineNumber(frame, lineno)  (frame)->f_lineno = (lineno)
            |                                                              ^~
      IT8951/spi.c:22091:5: note: in expansion of macro ‘__Pyx_PyFrame_SetLineNumber’
      22091 |     __Pyx_PyFrame_SetLineNumber(*frame, firstlineno);
            |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      IT8951/spi.c: In function ‘__Pyx_PyBytes_Equals’:
      IT8951/spi.c:23312:13: warning: ‘ob_shash’ is deprecated [-Wdeprecated-declarations]
      23312 |             hash1 = ((PyBytesObject*)s1)->ob_shash;
            |             ^~~~~
      In file included from /usr/include/python3.11/bytesobject.h:62,
                       from /usr/include/python3.11/Python.h:50,
                       from IT8951/spi.c:6:
      /usr/include/python3.11/cpython/bytesobject.h:7:35: note: declared here
          7 |     Py_DEPRECATED(3.11) Py_hash_t ob_shash;
            |                                   ^~~~~~~~
      IT8951/spi.c:23313:13: warning: ‘ob_shash’ is deprecated [-Wdeprecated-declarations]
      23313 |             hash2 = ((PyBytesObject*)s2)->ob_shash;
            |             ^~~~~
      /usr/include/python3.11/cpython/bytesobject.h:7:35: note: declared here
          7 |     Py_DEPRECATED(3.11) Py_hash_t ob_shash;
            |                                   ^~~~~~~~
      IT8951/spi.c: In function ‘__Pyx_AddTraceback’:
      IT8951/spi.c:438:62: error: invalid use of incomplete typedef ‘PyFrameObject’ {aka ‘struct _frame’}
        438 |   #define __Pyx_PyFrame_SetLineNumber(frame, lineno)  (frame)->f_lineno = (lineno)
            |                                                              ^~
      IT8951/spi.c:24384:5: note: in expansion of macro ‘__Pyx_PyFrame_SetLineNumber’
      24384 |     __Pyx_PyFrame_SetLineNumber(py_frame, py_line);
            |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      error: command '/usr/bin/aarch64-linux-gnu-gcc' failed with exit code 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: legacy-install-failure

× Encountered error while trying to install package.
╰─> IT8951

note: This is an issue with the package mentioned above, not pip.
hint: See above for output from the failure.`

JonCellini:
That IT8951 issue appears to be a dupe of what was reported earlier [here](https://github.com/TomWhitwell/SlowMovie/issues/150#issuecomment-1766641371) .

JonCellini:
Yep, I just came across the same fix and tested it on my Bookworm install that I was testing this PR against, and it builds cleanly. The workaround is a very simple change to the installer which I'm happy to commit as well.

JonCellini:
Latest commit addresses issue [#152](https://github.com/TomWhitwell/SlowMovie/issues/152).

JonCellini:
I did some testing this evening on a totally fresh image and found an error which I resolved. A fresh install should build cleanly in a venv on Buster

JonCellini:
So I think I tracked down the issue with waveshare_epd working on Bookworm. They are looking for `gpiomem-bcm2835` which appears to have been replaced with `raspberrypi-gpiomem`. 

I did a clean install of my code on a fresh Bookworm system and it works as expected; however, the waveshare_epd will fail to run as is it is hard coded to look for gpiomem-bcm2835 to determine it's on a PI. That is over my depth of understanding to resolve at the moment.

robweber:
Thanks for all the work on this. Going to get Bookworm going on an SD card and see what I can figure out for the waveshare stuff.

robweber:
I put more info in #150 but just a note that we'll put a hold on this until the Bookworm issues in the Waveshare library can be resolved since that's a big chunk of the user base. For now using Bullseye is probably the best bet.

Install/install.sh
```diff
@@ -16,11 +16,20 @@ SERVICE_FILE_TEMPLATE=slowmovie.service.template
...
+function create_python_venv(){
+  python3 -m venv --system-site-packages $LOCAL_DIR/.venv # numpty requires system-site-packages
+  source $LOCAL_DIR/.venv/bin/activate #activate our venv
 }
 
 function install_python_packages(){
-  pip3 install -r $LOCAL_DIR/Install/requirements.txt -U
+  # fix PEP 517 error with OmniEPD
+  pip3 install --upgrade pip setuptools wheel
+  #fix for IT8951 install https://github.com/GregDMeyer/IT8951/issues/58
```
robweber, line 30: This shouldn't be needed anymore with the latest IT8951 repo fixes. I tested building on both Bullseye and Bookworm and both installed without error now

Install/requirements.txt
```diff
@@ -1,3 +1,4 @@
+Cython==3.0.4
```
robweber, line 1: Can remove this as well, this is defined in the IT8951 repo

README.md
```diff
@@ -62,20 +62,23 @@ On the Raspberry Pi:
 1. Set up environment
    * Update package sources: `sudo apt update`
    * Make sure git is installed: `sudo apt install git`
-   * Make sure pip is installed: `sudo apt install python3-pip`
-2. Install Waveshare e-paper drivers
+   * Make sure pip is installed: `sudo apt install python3-pip python3-venv`
+2. Create & activate python virtual environment (venv)
+   * Create virtual environment: `python3 -m venv .venv`
```
robweber, line 67: when creating virtual environment need the `--system-site-packages` directive like in the install script

helloworld.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!${HOME}/SlowMovie/.venv/bin/python3
```
robweber, line 1: Thoughts on just removing this entirely? I'm 50/50. If running within the virtual environment with `python3 helloworld.py` it will ignore this line anyway. My hesitation is that we don't know the target filesystem, what if the user decided to install this in `${HOME}/projects/SlowMovie`? Also not sure it will expand ${HOME} properly, it didn't seem to work when I tested it.

robweber:
I added a few comments based on the changes to the IT8951 branch since it builds properly now.

Install/requirements.txt
JonCellini: Agreed. That shouldn't be needed anymore since it's been fixed upstream.

Install/install.sh
JonCellini: Agreed. I just tested it locally and with the upstream changes to IT8951 those workarounds aren't necessary anymore.

README.md
JonCellini: You're correct, it's in the actual function I added to the installer was just a miss on the README.md changes to do it manually.

helloworld.py
JonCellini: So I gave this some thought and I agree reverting that change is the way to go. My original intent was to make it easy for someone to use (so have it work out of the box) and avoid the need to source the virtual environment if they were running it interactively without the included service file. I think perhaps a better approach would be to just have the installer add the activation of the venv to the user profile so it happens automatically on login. This would cover the interactive use case w/o the need for a bunch of complex code to test and source the venv if it wasn't sourced. The other option I can think of is to write a small wrapper that would handle sourcing it and instruct people to use that vs. calling `python slowmovie.py`.

Thoughts on if either of those approaches are desirable? Maybe I'm over thinking this and it's better to expect that the end-user can handle sourcing the venv on their own.

JonCellini:
I went ahead and revised my fork to incorporate the changes discussed above based on the changes to the IT8951 upstream. Let me know what you think makes sense to do about handling the venv sourcing.

helloworld.py
robweber: I think in _most_ cases it's probably correct to think that setting up SlowMovie will be pretty much the only application for the Pi. There is probably the off case - like testing - where it isn't but most people want to set it up and display it somewhere. I think if you can add activating the environment to the profile that would work. More DIY users will know how to remove it or are installing manually anyway.

JonCellini: Cool, I agree in most cases that would be what people do and making the default behavior "just work" would be the right thing to do for users. 

I went ahead and revised the installer to add logic to modify the installing users .profile file to handle sourcing the venv automatically on login. On a stock Raspberry PI OS system this is the file sourced for user logins - I figure if a DYI user has modified their shell from the OS default they are probably capable of handling sourcing it themselves.

The only other thing I can think of changing is a cosmetic thing. Right now the venv is called "venv" so once it's sourced the user shell looks like this:

_(.venv) jon@CodeVM:~$_ 

Since the project is named **SlowMovie** would it make more sense to call it "SlowMovie" so it looked like this instead?

_(.SlowMovie) jon@CodeVM:~$_

It would be a simple change to make but I'd like your input on what it should be.

robweber: I think it's a nice touch so it doesn't look so generic. Getting this up and running should be as non-technical as possible so anything that can help with that is good.

JonCellini: Done & tested. My latest commit changes it to be "**SlowMovie**"

JonCellini:
Let me know if you think any other changes are needed. I tested it locally and everything works as I expect.

.gitignore
```diff
@@ -75,3 +75,5 @@ Videos/*
 !Videos/test.mp4
 slowmovie.log
 slowmovie.conf
+# python venv
+.venv
```
robweber, line 79: should this be `.SlowMovie` now?

JonCellini: Yep, I overlooked changing it there. I just pushed a correction.

PR #155, opened 2023-11-13
Omni-epd 0.4.0 for Bookworm

robweber:
This updates the omni-epd library to the latest version, compatible with Raspberry Pi OS 12 (Bookworm). See the [release details ](https://github.com/robweber/omni-epd/releases/tag/v0.4.0)for a full list of changes. Mostly these incorporate updates to the IT8950 repo and Waveshare packages so they'll run on both Bullseye and Bookworm systems. 

This, coupled with #150 make SlowMovie fully compatible with both Bullseye and Bookworm installs now. I also removed redundant references to install the Waveshare drivers as the current Waveshare repo will not work on Bookworm. Right now a fork is used in omni-epd, which will be reverted back to the official repo once it's compatible.

### DIFF
--- a/Install/requirements.txt
+++ b/Install/requirements.txt
@@ -1,5 +1,4 @@
 ffmpeg-python==0.2.0
 Pillow>=9.1.0
 ConfigArgParse==1.4.1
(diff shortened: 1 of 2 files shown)


robweber:
I usually let these sit for a bit before merging to make sure there are no issues but don't want to hold this up too long. Any issues after merge please make an Issue.

PR #156, opened 2023-11-28
Update omni-epd to 0.4.1

robweber:
Updates the omni-epd version to the most current. This version reverts the Waveshare repo location back to the official source now that support for Bookworm and Raspberry Pi 5 is in the main branch there.


robweber:
Closing this. Subsequent tests have found (more) issues in the official Waveshare repo. Best to keep things where they are for now.

